PR #3, opened 2017-12-11
[QA] Upgraded to django 2.0

AZtheAsian:


requirements.txt
```diff
@@ -1,4 +1,3 @@
-django>=1.11,<2.0
```
nemesisdesign: I don't think is a good idea to remove this, use instead `django>=1.11,<2.1`
the travis build should not be affected because we'll be installing django before this requirement file is processed and the django requirement will be already satisfied.

django_loci/channels/base.py
```diff
@@ -29,12 +29,20 @@ def connect(self, message, pk):
...
+                    user.is_staff and
+                    user.has_perm(perm)
+                )
+            )
+        except TypeError:
+            return user.is_authenticated and (
+                user.is_superuser or (
+                    user.is_staff and
+                    user.has_perm(perm)
+                )
```
nemesisdesign: let's simplify this block of code and avoid duplication, try something like this:

```python
authenticated = user.is_authenticated
if callable(authenticated):
    authenticated = authenticated()
return authenticated and (
    # ... etcetera ...
)
```

django_loci/base/models.py
```diff
@@ -62,7 +62,7 @@ def short_type(self):
 
 @python_2_unicode_compatible
 class AbstractFloorPlan(TimeStampedEditableModel):
-    location = models.ForeignKey('django_loci.Location')
+    location = models.ForeignKey('django_loci.Location', on_delete=models.PROTECT)
```
nemesisdesign: the default in django 1.1 was `models.CASCADE` so changing it to protect would be a change in behaviour, which we are not aiming to do now.

Put `models.CASCADE` in place of `models.PROTECT`

django_loci/base/models.py
```diff
@@ -108,7 +108,7 @@ class AbstractObjectLocation(TimeStampedEditableModel):
         ('indoor', _('Indoor')),
         ('mobile', _('Mobile')),
     )
-    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    content_type = models.ForeignKey(ContentType, on_delete=models.PROTECT)
```
nemesisdesign: can we avoid changing this line please?

.travis.yml
```diff
@@ -22,6 +22,8 @@ branches:
 before_install:
   - pip install -U pip wheel setuptools
   - pip install --no-cache-dir -U -r requirements-test.txt
+  - if [[ $TRAVIS_PYTHON_VERSION == "3.6" ]]; then pip install "django>=2.0,<2.1"; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == "2.7" ]]; then pip install "django>=1.11,<2.0"; fi
```
nemesisdesign: move these two lines to the beginning of the `install` section of `.travis.yml` in order to not slow down the style checks which need to be performed very fast

nemesisdesign:
Oh man I forgot to merge this one.

PR #6, opened 2018-01-02
[travis] Add JSLint check

phyxkal:


.travis.yml
```diff
@@ -32,6 +33,7 @@ install:
   - if [[ $TRAVIS_PYTHON_VERSION == "2.7" ]]; then pip install git+git://github.com/tinio/pysqlite.git@extension-enabled#egg=pysqlite; fi
 
 script:
+  - jslint django_loci/static/django-loci/js/*js
```
nemesisdesign: move this to the end of `before_install`

django_loci/static/django-loci/js/floorplan-widget.js
```diff
@@ -31,7 +31,7 @@
         map.setView([0, 0], 0);
 
         function updateInput(e) {
-            var latlng = e.latlng || e.target._latlng;
+            var latlng = e.latlng || e.target.latlng;
```
nemesisdesign: not sure this is ok, does leaflet offer this attribute? If yes please show reference to existing practices/docs
Have you tested it?

phyxkal: so this should not be changed, but add a new configuration to .jslintrc?

PR #9, opened 2018-09-02
Channels 2 upgrade

### LINKED ISSUE #7: Upgrade to latest django-channels version
Upgrade to latest django-channels version.

nemesisdesign:
Thank you @atb00ker for your outstanding work! I've fixed the test and performed clean up.
Let's see if the travis build passes.

Closes #7 and #8.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Byte-compiled / optimized / DLL files
 __pycache__/
+.pytest_cache/
 *.py[cod]
 
 # C extensions
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ addons:
 
 python:
   - "3.6"
-  - "2.7"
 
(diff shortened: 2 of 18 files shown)


atb00ker:
@nemesisdesign , the file `pytest_channels.py` needs to be renamed to back `test_channels.py`.
I changed the name because i was running the file in a different manner! Currently that file is not running!

nemesisdesign:
@atb00ker oh I didn't notice that.. I already changed the logic again.. but anyway, I think now it's better, the result is cleaner and I undertsood how to make sure the `pytest` suite runs fast (it was running very slowly in my env because the db was being recreated at each test).

What do you think? Can we merge?

nemesisdesign:
PS: coveralls now reports 100%, not sure why it hasn't posted the result here: https://coveralls.io/builds/18793941

PR #11, opened 2018-11-22
[qa] Geometry shouldn't be allowed to be None if not mobile

### LINKED ISSUE #10: Geometry shouldn't be allowed to be None if not mobile
`Location.geometry` can be NULL, but that should be allowed only `if_mobile` is `True`, otherwise the model should raise a `ValidationError`.

cking100:
Fixes #10


cking100:
@nemesisdesign @atb00ker I'm not sure if i did it right or not please have a look and lemme know.
Thanks

django_loci/base/admin.py
```diff
@@ -230,7 +230,10 @@ def _get_location_instance(self):
         location.name = data.get('name') or location.name
         location.address = data.get('address') or location.address
         location.geometry = data.get('geometry') or location.geometry
-        return location
+        if not location.is_mobile and not location.geometry:
+            raise ValidationError(_('Geometry can\'t be null'))
```
nemesisdesign: move this logic to the base model, that's where it belongs, the admin is just a view, the main logic (also known as business logic) goes into the model, at least in django apps this is how we do it

cking100: Okay.. @nemesisdesign i suppose i have to put it here https://github.com/openwisp/django-loci/blob/1c5ed122f526e3ba7f88b27cb67f4010d7493e95/django_loci/base/models.py#L20 right? If yes then do i have to make a separate function to check this?

cking100:
@nemesisdesign Please have a look now and lemme know.

django_loci/base/models.py
```diff
@@ -54,6 +54,12 @@ def _validate_outdoor_floorplans(self):
             msg = 'this location has floorplans associated to it, ' \
                   'please delete them before changing its type'
             raise ValidationError({'type': msg})
+    
+    def _validate_geometry_if_not_mobile(self):
+        if not self.is_mobile and not self.geometry:
+            raise ValidationError(_('Geometry cannot be null'))
+        else:
+            return
```
nemesisdesign: remove the previous 2 lines

django_loci/base/models.py
```diff
@@ -54,6 +54,12 @@ def _validate_outdoor_floorplans(self):
             msg = 'this location has floorplans associated to it, ' \
                   'please delete them before changing its type'
             raise ValidationError({'type': msg})
+    
+    def _validate_geometry_if_not_mobile(self):
```
nemesisdesign: you should call this method from `clean`

django_loci/tests/base/test_models.py
```diff
@@ -71,6 +71,15 @@ def test_floorplan_association_validation(self):
...
             self.fail('ValidationError not raised')
 
+    def test_geometry_if_not_mobile(self):
+        try:
+            self._create_location(geometry=None)
+        except ValidationError as e:
+            err_str = str(e)
+            self.assertIn('geometry cannot be null', err_str)
+        else:
+            self.fail('ValidationError not raised')
```
nemesisdesign: Great. Please also add another test method which does the following:

```python
try:
    self._create_location(is_mobile=True, geometry=None)
except ValidationError:
    self.fail('Unexpected validation error raised')
```

cking100:
Done! @nemesisdesign  Please have a look :)

PR #21, opened 2019-08-14
[models] checked if self in _validate_location_type() has location attribute

oltarzewskik:


oltarzewskik:
`travis` throw this error:
firste
after changing `requirements` throw that:
seconde
What should we do in that situation?

django_loci/base/models.py
```diff
@@ -97,10 +97,12 @@ def delete(self, *args, **kwargs):
         self._remove_image()
 
     def _validate_location_type(self):
-        if self.location and self.location.type != 'indoor':
-            msg = 'floorplans can only be associated '\
-                  'to locations of type "indoor"'
-            raise ValidationError(msg)
+        if hasattr(self, 'location'):
```
nemesisdesign: please avoid nested ifs.

rewrite this as:

```python
if not hasattr(self, 'location') or not hasattr(self.location, 'type'):
    return
# rest of code
```

nemesisdesign:
One question @oltarzewskik, does this PR address one of the open issues we have in this repo?

oltarzewskik:
@nemesisdesign No, it is related to this: https://github.com/openwisp/openwisp-controller/pull/118

PR #22, opened 2019-08-19
[enchancment] Updated requirements and temporary travis build fix

oltarzewskik:


.travis.yml
```diff
@@ -29,7 +30,8 @@ before_install:
 
 install:
   - python setup.py -q develop
-
+  # temp build fix
+  - pip install django==1.11.23
```
nemesisdesign: I don't understand this line

oltarzewskik: To avoid travis error (in that PR: https://github.com/openwisp/django-loci/pull/21) I had to use older version of ubuntu (trusty), this ubuntu distro has deprecated `sqlite` packages and they weren't working with `Django 2.*` versions therefore I overread installed via requirement `django` version in travis to not mess up requirements

nemesisdesign: can you try removing the lines 33 and 34 to see if the build works?
The line `pip install django==1.11.23` is not acceptable because it binds the buid to an old version of django.

oltarzewskik: In my comment above I meant `Django 2.1.*` or higher instead of `Django 2.*` , my bad

I had to change version of ubuntu to `trusty` because in other way (if version was newer) `travis` built would throw that error: https://github.com/openwisp/django-loci/pull/21#issuecomment-521583444 to avoid this I added `dist: trusty` in `travis` (the same thing is done in `openwisp-controller`, in `controller` I already have added issue for adding support for newer version of `ubuntu` in `travis`, should I add the same in other repos where `trusty` is used (`django-loci` obviously too) )

, but there is bad side of that namely in `Django 2.1` `django` dropped support for sqlite version used by `trusty` (and BTW it is not that easy to update the version of sqlite) therefore I had to install `Django` < `2.1` because in other way `travis will throw error`

oltarzewskik: and the error which is occuring now in `travis` is throwing even when I undo all changes of the commit

requirements.txt
```diff
@@ -1,6 +1,6 @@
-django>=1.11,<2.2
 django-leaflet>=0.23,<0.25
-channels>=2.1.3,<2.2.0
-openwisp-utils>=0.2,<0.3.0
-Pillow>=4.3.0,<6.0
+channels>=2.2.0,<2.3.0
+Pillow>=4.3.0,<6.2
 openwisp-utils>=0.2.1,<0.4.0
+pyasn1==0.4.8
+django==2.0.13
```
oltarzewskik: @nemesisdesign `travis` is overwritting django anyway:
https://travis-ci.org/openwisp/django-loci/builds/618705500?utm_source=github_status&utm_medium=notification

nemesisdesign: please remove this

requirements.txt
```diff
@@ -1,6 +1,6 @@
-django>=1.11,<2.2
 django-leaflet>=0.23,<0.25
-channels>=2.1.3,<2.2.0
-openwisp-utils>=0.2,<0.3.0
-Pillow>=4.3.0,<6.0
+channels>=2.2.0,<2.3.0
+Pillow>=4.3.0,<6.2
 openwisp-utils>=0.2.1,<0.4.0
+pyasn1==0.4.8
```
nemesisdesign: shouldn't be this installed automatically?

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: python
 sudo: false
 cache: pip
+dist: trusty
```
nemesisdesign: can you tell me again why we need to set this distro and we can't use a more recent one?

requirements.txt
oltarzewskik: It should but it isn't

.travis.yml
oltarzewskik: it's because otherwise `travis` would show error like here: https://github.com/openwisp/django-loci/pull/21#issuecomment-521583444

nemesisdesign:
I fixed the build in a way that I think is cleaner. The distro is more up to date and the solution should be more robust.
See my changes here: https://github.com/openwisp/django-loci/pull/27

Issue #24, opened 2019-11-21
[build] Test multiple django and python versions


nemesisdesign:
In the travis build that runs automated tests, we should be testing the different supported django versions and at least python 3.6 and python 3.7.

PS: we can drop support for django 1.11.x.

**Reserved for GCI**.

PR #31, opened 2019-12-05
[build] Test multiple django and python versions

Weirdo914:
Added Support for python 3.7
Dropped Support for django 1.11
Fixes https://github.com/openwisp/django-loci/issues/24

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 django>=1.11,<2.2
 django-leaflet>=0.23,<0.25
-channels>=2.1.3,<2.2.0
-openwisp-utils>=0.2,<0.3.0
-Pillow>=4.3.0,<6.0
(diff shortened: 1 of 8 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-django>=1.11,<2.2
+django>=2.0,<2.3
```
atb00ker: Hi,

Why are you dropping `1.11`?
For one, it supports python 3.7 and second changing the requirements is not a part of this task.

requirements.txt
```diff
@@ -1,4 +1,3 @@
-django>=1.11,<2.2
```
atb00ker: Wait, why did you remove django from requirements, it was fine before!
Since we discussed on gitter that `1.11` can be dropped. `django>=2.0,<2.3` would be fine!

PR #33, opened 2019-12-08
[ux] Fix loci fields hidden when error occurs

### LINKED ISSUE #38: [ux] errors made while adding a new device gets hidden
When adding a new device if someone makes an error of not entering `Geometry` for a `Outdoor environment`, the user will get error `Please correct the errors below.` at the top of the page but the actual field to make the correction is not shown.
The user needs to change the `Type` field then select `Outdoor environment` again to see the errored field.
This may cause confusion to users.

marfgold1:
`$geoRows.hide()` causing the loci coords fields to hidden by default, even if error occurs.
It's because *initially*, locationSelectionChange will **not** reset the form, so when user already set the `Location selection` and `Type` field then try to save it, if there's an error occurs, then `.loci.coords .form-row` fields will be hidden nevertheless because this `$geoRows.hide()` calls, even though the fields should not be hidden. This of course can confuse users because the UI inconsistency.

Closes #38

### DIFF
--- a/django_loci/static/django-loci/js/loci.js
+++ b/django_loci/static/django-loci/js/loci.js
@@ -9,7 +9,6 @@ django.jQuery(function ($) {
     var $outdoor = $('.loci.coords'),
         $indoor = $('.indoor.coords'),
         $allSections = $('.coords'),
-        $geoRows = $('.loci.coords .form-row'),
         $geoEdit = $('.field-name, .field-type, .field-is_mobile, ' +
                      '.field-address, .field-geometry', '.loci.coords'),
         $indoorRows = $('.indoor.coords .form-row:not(.field-indoor)'),
@@ -119,12 +118,10 @@ django.jQuery(function ($) {
         if (!initial) { resetDeviceLocationForm(); }
         if (value === 'new') {
             $outdoor.show();
-            $geoRows.hide();
             $typeRow.show();
             indoorForm(value);
         } else if (value === 'existing') {
             $outdoor.show();
-            $geoRows.hide();
             $locationRow.show();
         }
     }


marfgold1:
Done @nemesisdesign  👍 
Also don't forget to review my issues 😄
Thank you very much 😃

marfgold1:
@atb00ker @nemesisdesign Also the reason why I also deleted the $geoRows when Location selection is on both condition (Existing and New) because $geoRows.hide() is hiding the field, but there was a check that if **location is defined** (in another word, location selection is Existing and the user already set the Location), then `$geoEdit.show()` was called **after** the locationSelectionChange() called, showing the fields again so it's look fine.
Not like when locationSelection is on New, **no calls will show the fields again** (because location was not defined since locationSelection is New).
That's why, in conclusion, $geoRows should've been removed anyway because `typeChange()` is already done a good job hiding and showing the fields.

Here is my research:
A. When Location selection is New, just when I save the device
1. Before typeChange call
Screenshot from 2019-12-09 19-55-31
2. After typeChange call
Screenshot from 2019-12-09 19-55-35
3.a. When locationSelectionChange is called, before $geoRows.hide() call
Screenshot from 2019-12-09 19-48-32
3.b. When locationSelectionChange is called, after $geoRows.hide() call
Screenshot from 2019-12-09 19-48-44
4 After locationSelectionChange is called
Screenshot from 2019-12-09 19-55-51
5. After the script fully loaded
Screenshot from 2019-12-09 19-48-59

B. When Location selection is Existing
1. Before typeChange call
Screenshot from 2019-12-09 20-04-12
2. After typeChange call
Screenshot from 2019-12-09 20-04-27
3. When locationSelectionChange is called, after $geoRows.hide() call
Screenshot from 2019-12-09 20-04-40
4. After $location.val() check
Screenshot from 2019-12-09 20-06-45
5. After the script fully loaded
Screenshot from 2019-12-09 20-05-03

Yes, the time was not in order, but don't worry, it's still have the same value and code :sweat_smile:. I hope the images will clear the confusion :wink:, but feel free to ask again :smile:. Sorry if it's still not explained properly :disappointed:

PR #39, opened 2019-12-08
[bug] Fixed frontend bug

atb00ker:
Fixed frontend bug introduced in Django 2.2 update

### DIFF
--- a/django_loci/static/django-loci/js/floorplan-widget.js
+++ b/django_loci/static/django-loci/js/floorplan-widget.js
@@ -1,7 +1,7 @@
-(function ($) {
+(function () {
     'use strict';
     django.loadFloorPlan = function (widgetName, imageUrl, imageW, imageH) {
-        var $input = $('#id_' + widgetName),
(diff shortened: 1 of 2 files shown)


django_loci/static/django-loci/js/floorplan-widget.js
```diff
@@ -1,7 +1,7 @@
-(function ($) {
+(function () {
     'use strict';
     django.loadFloorPlan = function (widgetName, imageUrl, imageW, imageH) {
-        var $input = $('#id_' + widgetName),
+        var $input = django.jQuery('#id_' + widgetName),
```
atb00ker: This was the cause of the problem.
```
Aliasing the jQuery Object
When $.noConflict() is used to avoid namespace conflicts, the $ shortcut is no longer available. 
```
Hence `$` is not a local variable when this script runs for the first time on page load. 
Our options are using `django.jQuery` or passing `$` as a arguement to this function on call.

I used `django.jQuery` and tested it, it's working!

nemesisdesign:
Thanks! :+1:

PR #42, opened 2020-01-02
[qa] Replace with openwisp-utils-qa-checks

waleko:
Replace old qa checks with `openwisp-utils-qa-checks` (related to openwisp/openwisp-utils#63 for this repo)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,9 +30,6 @@ before_install:
   - pip install -U pip wheel setuptools
   - pip install -U $DJANGO
   - pip install --no-cache-dir -U -r requirements-test.txt
-  - ./runflake8
-  - ./runcheckmigration
-  - ./runisort
   - npm install -g jslint
   - jslint django_loci/static/django-loci/js/*.js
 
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -30,9 +30,7 @@ before_install:
   - pip install -U pip wheel setuptools
   - pip install -U $DJANGO
   - pip install --no-cache-dir -U -r requirements-test.txt
-  - ./runflake8
```
atb00ker: Please delete these files as well! :smile:

.travis.yml
```diff
@@ -30,9 +30,7 @@ before_install:
...
-  - ./runcheckmigration
-  - ./runisort
+  - openwisp-utils-qa-checks --migration-path "./django_loci/migrations ./django_loci/tests/testdeviceapp/migrations" --skip-checkmakemigrations
```
nemesisdesign: please use a multiline statement for better readability, ensure you do this in all similar PRs

.travis.yml
```diff
@@ -30,9 +30,12 @@ before_install:
...
   - pip install --no-cache-dir -U -r requirements-test.txt
-  - ./runflake8
-  - ./runcheckmigration
-  - ./runisort
+  - |
+      openwisp-utils-qa-checks \
+      --migration-path \
+        "./django_loci/migrations \
+        ./django_loci/tests/testdeviceapp/migrations" \
+      --skip-checkmakemigrations
```
nemesisdesign: why are you using the `skip-checkmakemigrations` flag?

waleko: Very good question... I'll fix it right away

.travis.yml
```diff
@@ -30,26 +30,22 @@ before_install:
...
   - pip install --no-cache-dir -U -r requirements-test.txt
-  - ./runflake8
-  - ./runcheckmigration
-  - ./runisort
   - npm install -g jslint
   - jslint django_loci/static/django-loci/js/*.js
 
 install:
   - pip install -e .
+  - python setup.py -q develop
```
nemesisdesign: this is not necessary (`pip install -e .` is the equivalent using pip), why did you add it? I only spotted it now

waleko: @nemesisdesign Otherwise `django_loci` module isn't set-up, so makemigrations check fails

nemesisdesign: @waleko `pip install -e .` is equivalent to `python setup.py -q develop`, so we're doing twice the same operation, if without this additional line it fails it's likely because some other reason which we would be covering this way

waleko: @nemesisdesign My bad... I took a one more look at it and it turned out, that I've put `openwisp-utils-qa-checks` before `pip install -e .`

Issue #44, opened 2020-01-08
[ux] Automatically calculate geographic coordinates or address


nemesisdesign:
When the position on the geographic map is updated:

1. if the address field is empty we should automatically get the best address from the geographic coordinates and update the address field with the result
2. if the address field is not empty we should ask the user if the user wants the map to be updated (for now with a simple javascript confirm dialog); If the user replies affirmatively we can do the same described in point 1

When address field is updated:

1. If the geographic coordinates on the map are not set, we can get the geographic coordinates  from the address and update the map accordingly
2 if the geographic coordinates on the map are already set, we should ask the user if the user wants the map to be updated (for now with a simple javascript confirm dialog); If the user replies affirmatively we can do the same described in point 1

For the geocoding operations, we can use [geopy](https://geopy.readthedocs.io/).

We should implement an abstraction that allows to change the provider, trying to support at least Google Maps, Nominatim and ArcGIS in the first version fo this feature.

We should handle failures and retry, retry by default 3 times with a delay of 1 second in case of failures, the geocoding library should have a built-in implementation for this, [see RateLimiter](https://github.com/geopy/geopy/blob/5f8e457b700b59221401726e5983175156b20a0a/geopy/extra/rate_limiter.py#L15).

We will likely need the addition of a few django global settings that allow to set:
- the chosen provider, defaults to `ArcGIS`
- the API key, defaults to `None`
- whether the django app may refuse to start if the testing of the geocoding operation fails, defaults to `True`, eg: `DJANGO_LOCI_STRICT_GEOCODE_TEST`
- amount of retries after failures, default to 3
- delay in seconds before trying again after a failure, default to 1

On startup, the app shall test an arbitrary geocoding operation (eg: get the coordinates of the address of the Python Foundation) and if it fails it will raise an `ImproperlyConfigured` exception if ``DJANGO_LOCI_STRICT_GEOCODE_TEST`` is `True` or it will log an error otherwise (using the python logging facility).

Since this app does not use django-rest-framework, the geocoding operations should be performed via a [simple JSON view API that it is added to the admin as we do in django-netjsonconfig](https://github.com/openwisp/django-netjsonconfig/blob/3d14334391f99e964b85826fe085ea0b8a99b6e6/django_netjsonconfig/base/admin.py#L84-L98) (see `context_view` as a reference), although the logic to perform the geocoding can be stored in a function or a class to improve reusability.

This feature needs to be documented in the README and good tests must be written for it.

PR #46, opened 2020-01-12
[ux] Automatically calculate geodata #44

oltarzewskik:


rohit-png:
how to give a pull request

oltarzewskik:
@rohit-png checkout: https://help.github.com/en/desktop/contributing-to-projects/creating-a-pull-request

oltarzewskik:
JS part is not ready yet, however is enough for python part testing, please focus on python part of code (if somebody is interested in testing via current JS please let me know)
- I haven't added yet support for google maps, but `nominatim`, and `ArcGIS` seems to work (more pragmatic tests I will do tomorrow)
- error via logger is outputting `ERROR:root:Geocode test failed`, should I format it?

django_loci/geo/geocode.py
```diff
@@ -0,0 +1,41 @@
...
+        key = request.GET.get('key', None)
+        not_found = False
+        if key == DJANGO_LOCI_GEOCODE_API_KEY:
+            if address:
+                location = self.geocode(address)
+                if location != None:
+                    lat = location.latitude
+                    lng = location.longitude
+                else:
+                    not_found = True
```
oltarzewskik: This line will be used in future to output message in UI if location haven't been found

django_loci/admin.py
```diff
@@ -39,6 +41,12 @@ class ObjectLocationInline(AbstractObjectLocationInline):
     model = ObjectLocation
     form = ObjectLocationForm
 
+_admin_site_get_urls = admin.site.get_urls
 
+def get_urls():
+    return _admin_site_get_urls() + [url(r'^geocode/$', admin.site.admin_view(GeoCodeView.as_view()))]
```
nemesisdesign: why are you putting this hear instead of `base.admin`, also, why can't you override the `get_urls` as we do in the other openwisp-modules?

django_loci/apps.py
```diff
@@ -15,11 +23,21 @@ def __setmodels__(self):
         from .models import Location
         self.location_model = Location
 
+    def test_geocode(self):
+        location = GeoCodeView.geocode('Red Square')
+        try:
+            assert round(location.latitude) == 56 and round(location.longitude) == 38
+        except AssertionError:
+            if DJANGO_LOCI_GEOCODE_STRICT_TEST:
+                raise ImproperlyConfigured
```
nemesisdesign: `raise ImproperlyConfigured('Geocode test failed')`

django_loci/apps.py
```diff
@@ -15,11 +23,21 @@ def __setmodels__(self):
         from .models import Location
         self.location_model = Location
 
+    def test_geocode(self):
+        location = GeoCodeView.geocode('Red Square')
+        try:
+            assert round(location.latitude) == 56 and round(location.longitude) == 38
+        except AssertionError:
```
nemesisdesign: I would expect it to fail for other reasons, not `AssertionError`. I would expect it to fail for a network issue, the API key is incorrect, things like that. Have you tried to disconnect your laptop from the internet and run this and see what happens?

django_loci/geo/geocode.py
```diff
@@ -0,0 +1,41 @@
...
+from django.views import View
+from django.utils.module_loading import import_string
+
+from .settings import DJANGO_LOCI_GEOCODER, DJANGO_LOCI_GEOCODE_API_KEY, DJANGO_LOCI_GEOCODE_DELAY, DJANGO_LOCI_GEOCODE_RETRIES
+
+from geopy.extra.rate_limiter import RateLimiter
+
+
+class GeoCodeView(View):
+    geocoder = import_string('geopy.geocoders.' + DJANGO_LOCI_GEOCODER)
```
nemesisdesign: learn to use `format`: `'geopy.geocoders.{}'.format(DJANGO_LOCI_GEOCODER)` which is more pythonic

django_loci/geo/settings.py
```diff
@@ -0,0 +1,21 @@
+from django.conf import settings
+
+DJANGO_LOCI_GEOCODER = getattr(settings,
+                               'DJANGO_LOCI_GEOCODER',
+                               'ArcGIS')
+
```
nemesisdesign: remove blank lines between each setting and write all the setting definition on one line if possible because they're pretty simple definition which do not need to be laid out on multiple lines (we do that for more complex data structures).

django_loci/geo/settings.py
```diff
@@ -0,0 +1,21 @@
...
+                                          'DJANGO_LOCI_STRICT_GEOCODE_TEST',
+                                          True)
+
+DJANGO_LOCI_GEOCODE_DELAY = getattr(settings,
+                                    'DJANGO_LOCI_GEOCODE_DELAY',
+                                    1)
+
+DJANGO_LOCI_GEOCODE_RETRIES = getattr(settings,
+                                      'DJANGO_LOCI_GEOCODE_RETRIES',
+                                      3)
```
nemesisdesign: ensure these settings are explained in the README

django_loci/static/django-loci/js/loci.js
```diff
@@ -365,6 +373,45 @@ django.jQuery(function ($) {
     // this is triggered in the location form page
     } else if (!$type.length) {
         var pk = window.location.pathname.split('/').slice('-3', '-2')[0];
+        var api = 'http://' + window.location.host + '/admin/geocode/';
```
nemesisdesign: do not harcode `http://`, use `window.location.protocol`.

tests/settings.py
```diff
@@ -90,6 +90,12 @@
     }
 ]
 
+DJANGO_LOCI_GEOCODER = 'ArcGIS'
+DJANGO_LOCI_GEOCODE_API_KEY = None
+DJANGO_LOCI_GEOCODE_STRICT_TEST = True
+DJANGO_LOCI_GEOCODE_DELAY = 1
+DJANGO_LOCI_GEOCODE_RETRIES = 3
+
```
nemesisdesign: these look like the default values, I think you can remove them from here since they don't add any value

django_loci/apps.py
```diff
@@ -15,11 +23,21 @@ def __setmodels__(self):
         from .models import Location
         self.location_model = Location
 
+    def test_geocode(self):
+        location = GeoCodeView.geocode('Red Square')
```
nemesisdesign: don't use the view. Move the logic which does geocoding out of the view in a function or a few functions or a class or whatever and call the same code from the view, as explained in the issue description

django_loci/admin.py
nemesisdesign: I think it would be cleaner to have one URL for geocoding and one for reverse geocoding

django_loci/geo/geocode.py
```diff
@@ -0,0 +1,41 @@
...
+    geolocator = geocoder(user_agent="django_loci")
+    geocode = RateLimiter(geolocator.geocode, DJANGO_LOCI_GEOCODE_DELAY, DJANGO_LOCI_GEOCODE_RETRIES)
+    reverse = RateLimiter(geolocator.reverse, DJANGO_LOCI_GEOCODE_DELAY, DJANGO_LOCI_GEOCODE_RETRIES)
+
+    def get(self, request):
+        address = request.GET.get('address', '')
+        lat = request.GET.get('lat', 0)
+        lng = request.GET.get('lng', 0)
+        key = request.GET.get('key', None)
```
nemesisdesign: remove this, we don't have to leak the API key to the browser, there's no need for that

django_loci/apps.py
oltarzewskik: `AssertionError` will be raised every time if geocodeing doesn't work including network issues and incorrect API key

oltarzewskik:
I am not able to properly test google maps, because literally after a few requests they say that I have to enable billing

All expected utilities of js part are working correctly except save button doesn't actually save changes done on map by the js :P

django_loci/admin.py
```diff
@@ -40,5 +42,15 @@ class ObjectLocationInline(AbstractObjectLocationInline):
     form = ObjectLocationForm
 
 
+_admin_site_get_urls = admin.site.get_urls
+
+
+def get_urls():
```
oltarzewskik: It is the most clean that I found to add these views. I if add these views to https://github.com/openwisp/django-loci/blob/master/django_loci/base/admin.py#L60 they will look like these `admin/django_loci/location/<my-part>` and that won't work because first parameter after `location/` is expected to be `pk`

atb00ker:
> I am not able to properly test google maps, because literally after a few requests they say that I have to enable billing

Regarding the testing with GoogleMaps,
I can provide you with a temporary API key to avoid you to incur charges.
If you want one, please let me know. 
I created a key for myself and enabled `geocoding-backend.googleapis.com` API but it didn't work. Please let me know / and document any other steps as well! :-)

oltarzewskik:
> I am still testing but I am seeing:
> 
> ```
> Uncaught TypeError: Cannot read property '_layers' of undefined
>     at get_marker (loci.js:347)
>     at loci.js:405
> ```
> 
> on my console at add location page!

I figured out that for some reason this code doesn't work on firefox, if you have this error still anyway for example on chrom try to increase this number: https://github.com/openwisp/django-loci/pull/46/files#diff-522472a1463e32409d029072df76d951R455

oltarzewskik:
@atb00ker @nemesisdesign Now feature should work properly

django_loci/admin.py
nemesisdesign: I'm not sure of this, if we don't do the override in the base files we will have to repeat this code in openwisp-controller, it looks to me that you have to try harder.

Have you tried adding:

```python
url(r'^geocode/$', admin.site.admin_view(geocode_view)),
url(r'^reverse_geocode/$', admin.site.admin_view(reverse_geocode_view))
```

to this list? https://github.com/openwisp/django-loci/blob/master/django_loci/base/admin.py#L65

django_loci/static/django-loci/js/loci.js
```diff
@@ -377,4 +401,54 @@ django.jQuery(function ($) {
...
+                                getMap().setView([data['lat'], data['lng']])
+                            }
+                        }
+                        var wkt = '{ "type": "Point", "coordinates": [ ' + data['lng'].toString() + ', ' + data['lat'].toString() + '] }';
+                        $('.django-leaflet-raw-textarea').val(wkt);
+                    } else {
+                        alert('Not found location with given address!');
+                    }
+                    address_lat = data['lat']
+                    address_lng = data['lng']
```
nemesisdesign: use the JS linter to qa your code

django_loci/geo/geocode.py
```diff
@@ -0,0 +1,48 @@
...
+    if address:
+        location = geocode(address)
+        if location is not None:
+            lat = location.latitude
+            lng = location.longitude
+    response = {
+        'lat': lat,
+        'lng': lng,
+    }
+    return JsonResponse(response)
```
nemesisdesign: shorten to 
```python
return JsonResponse({
    'lat': lat,
    'lng': lng,
})

File: django_loci/geo/geocode.py [hunk 15]
```diff
@@ -0,0 +1,48 @@
...
+    lat = request.GET.get('lat', 0)
+    lng = request.GET.get('lng', 0)
+    if lat and lng:
+        location = reverse_geocode((lat, lng))
+        if location is not None:
+            address = location.address
+    response = {
+        'address': address
+    }
+    return JsonResponse(response)
```
nemesisdesign: same here, simplify, make your code more compact

README.rst
```diff
@@ -162,6 +162,67 @@ To understand the details of this statement, take a look at the code of
...
+| **type**:    | ``str``     |
++--------------+-------------+
+| **default**: | ``ArcGIS``  |
++--------------+-------------+
+
+Service used for geocoding and reverse geocoding.
+
+Supported geolocation services:
+
+* `ArcGIS`
```
nemesisdesign: should be:
```
``ArcGIS``
# There are two ``
```

Same for the others

README.rst
```diff
@@ -162,6 +162,67 @@ To understand the details of this statement, take a look at the code of
...
+``DJANGO_LOCI_GEOCODE_STRICT_TEST``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-----------+
+| **type**:    | ``bool``  |
++--------------+-----------+
+| **default**: | ``True``  |
++--------------+-----------+
+
+If geocoding tests failed and set to ``False`` won't raise error.
```
nemesisdesign: Change to:

```
Indicates whether the system should raise an ``ImproperlyConfigured``
exception in case geocoding doesn't work when the application is started.
```

README.rst
```diff
@@ -162,6 +162,67 @@ To understand the details of this statement, take a look at the code of
...
+``DJANGO_LOCI_GEOCODE_FAILURE_DELAY``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+----------+
+| **type**:    | ``int``  |
++--------------+----------+
+| **default**: | ``1``    |
++--------------+----------+
+
+Amount of seconds between geocoding retry API calls if geocoding request failed.
```
nemesisdesign: `if geocoding request failed` > `when geocoding requests fail`

README.rst
```diff
@@ -162,6 +162,67 @@ To understand the details of this statement, take a look at the code of
...
+``DJANGO_LOCI_GEOCODE_RETRIES``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+----------+
+| **type**:    | ``int``  |
++--------------+----------+
+| **default**: | ``3``    |
++--------------+----------+
+
+Amount of retry API calls if geocoding request failed.
```
nemesisdesign: `if geocoding request failed` > `when geocoding requests fail`

README.rst
```diff
@@ -162,6 +162,67 @@ To understand the details of this statement, take a look at the code of
...
+``DJANGO_LOCI_GEOCODE_GOOGLEV3_API_KEY``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+-----------+
+| **type**:    | ``str``   |
++--------------+-----------+
+| **default**: | ``None``  |
++--------------+-----------+
+
+API key required Google Maps to perform geocoding requests.
```
nemesisdesign: `API key if required (eg: Google Maps)`

README.rst
```diff
@@ -162,6 +162,67 @@ To understand the details of this statement, take a look at the code of
...
+
++--------------+----------+
+| **type**:    | ``int``  |
++--------------+----------+
+| **default**: | ``3``    |
++--------------+----------+
+
+Amount of retry API calls if geocoding request failed.
+
+``DJANGO_LOCI_GEOCODE_GOOGLEV3_API_KEY``
```
nemesisdesign: rename this to `DJANGO_LOCI_GEOCODE_API_KEY`, we may add support for other services which also require an API key so this should be generic

django_loci/apps.py
```diff
@@ -15,11 +25,21 @@ def __setmodels__(self):
         from .models import Location
         self.location_model = Location
 
+    def test_geocoder(self):
+        try:
+            location = geocode('Red Square')
+            assert location is not None
+        except AssertionError:
+            if DJANGO_LOCI_GEOCODE_STRICT_TEST:
+                raise ImproperlyConfigured('Geocode utility is improperly configured')
```
nemesisdesign: this exception needs to show some more detailed information about what is going wrong so the user/developer can fix it

django_loci/apps.py
```diff
@@ -15,11 +25,21 @@ def __setmodels__(self):
...
+            self.logger.exception('Geocode test failed')
```
nemesisdesign: same here, I think this should output the same text as the line before, in the first case the app will be stopped and the exception raised, in this other case the exception will be logged but the application startup will not be halted

django_loci/apps.py
```diff
@@ -15,11 +25,21 @@ def __setmodels__(self):
...
+
     def ready(self):
         import leaflet
         leaflet.app_settings['NO_GLOBALS'] = False
         self.__setmodels__()
         self._load_receivers()
+        self.test_geocoder()
```
nemesisdesign: rename this method to `_test_geocoder`, it makes sense to mark it as private and is also consistent with the other methods

django_loci/admin.py
nemesisdesign: PS: use `reverse-geocode` instead of 'reverse_geocode` for the URL path, for consistency with the other OpenWISP modules in which we generally use dashes instead of underscores

django_loci/geo/geocode.py
```diff
@@ -0,0 +1,51 @@
...
+    })
+
+
+def reverse_geocode_view(request):
+    lat = request.GET.get('lat')
+    lng = request.GET.get('lng')
+    if lat and lng:
+        location = reverse_geocode((lat, lng))
+        if location is None:
+            address = ''
```
nemesisdesign: same 404 here

django_loci/geo/geocode.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def geocode_view(request):
+    address = request.GET.get('address')
+    if address is None:
+        return JsonResponse({'error': 'Address parameter not defined'}, status=400)
+    else:
+        location = geocode(address)
+        if location is None:
+            return JsonResponse({'error': 'Not found location with given name'}, status=404)
```
nemesisdesign: are you handling the 404 on the JS side?

django_loci/geo/geocode.py
oltarzewskik: Let's assume that geometry is on ocean, there is not address related to that geometry in that case maybe we should instead of 404 response just set address to empty string?

django_loci/geo/geocode.py
oltarzewskik: Yes, here: https://github.com/openwisp/django-loci/pull/46/files#diff-522472a1463e32409d029072df76d951R417
if response won't be 200 it have to be 404, because 400 response won't be raise because address will be always defined: https://github.com/openwisp/django-loci/pull/46/files#diff-522472a1463e32409d029072df76d951R382

oltarzewskik:
> Not filling the geometry field as expected:
try to create a new object
fill the address, press tab to let the JS place the point on the map
once the point appears on the map hit save (without futher editing)
Result: "this field is required for locations of type outdoor" appears in the geometry field row.

Could you send SS of that I am not able to reproduce it

oltarzewskik:
> Have you tried adding:

url(r'^geocode/$', admin.site.admin_view(geocode_view)),
url(r'^reverse_geocode/$', admin.site.admin_view(reverse_geocode_view))
to this list? https://github.com/openwisp/django-loci/blob/master/django_loci/base/admin.py#L65

Yes, the url looks like this: `admin/django_loci/location/geocode` which as I said before won't work, because first part after `/location/` must pk

requirements.txt
```diff
@@ -3,3 +3,4 @@ django-leaflet>=0.23,<0.27
 channels>=2.1.3,<2.5.0
 Pillow>=6.2.0,<=7.1.0
 openwisp-utils>=0.4.0,<0.5.0
+geopy>=1.16.0,<1.21.1
```
nemesisdesign: why are you using 1.16 minimum? If there's no specific reason use 1.20.0 as a minimum please.

django_loci/tests/base/test_geocode.py
```diff
@@ -0,0 +1,63 @@
+from django.urls import reverse
+
+from .. import TestAdminMixin
+
+
+class BaseTestGeocode(TestAdminMixin):
+    def test_geocode(self):
```
nemesisdesign: please move these tests to `test_admin` so that we avoid having to change openwisp-controller.

django_loci/static/django-loci/js/loci.js
```diff
@@ -343,6 +344,92 @@ django.jQuery(function ($) {
...
+                    var data_promise = resp.json();
+                    data_promise.then(function (data) {
+                        if (resp.status === 200) {
+                            var current_marker_lat = get_marker() !== undefined ?
+                                                        get_marker().getLatLng().lat : null;
+                            var current_marker_lng = get_marker() !== undefined ?
+                                                        get_marker().getLatLng().lng : null;
+
+                            if (current_marker_lat !== data.lat || current_marker_lng !== data.lng) {
+                                var wkt = '{ "type": "Point", "coordinates": [ ' +
```
nemesisdesign: why did you call this var `wkt`?
WKT is something like: `POINT (30 10)`. Maybe is better to call it geojson.

django_loci/static/django-loci/js/loci.js
```diff
@@ -377,4 +464,90 @@ django.jQuery(function ($) {
...
+                                if (current_marker_lat !== window.address_lat ||
+                                        current_marker_lng !== window.address_lng) {
+                                    if (resp.status === 200) {
+                                        if (window.confirm('Would you like to update address?')) {
+                                            $addressInput.val(data.address);
+                                            submit = true;
+                                            $('form').submit();
+                                        } else {
+                                            submit = true;
+                                            $('form').submit();
```
nemesisdesign: there's many calls to `$('form')` which could be stored in the beginning of the function as `$form= $('form')`.

django_loci/static/django-loci/js/loci.js
```diff
@@ -377,4 +464,90 @@ django.jQuery(function ($) {
...
+                                        }
+                                    } else {
+                                        var message = 'Current location does not have address, '
```
nemesisdesign: I think in this case we can simply let the user know that we didn't find any address, something like:

```
alert("We couldn't find a related address to the location selected on the map, please enter the address manually")
```

django_loci/static/django-loci/js/loci.js
```diff
@@ -377,4 +464,90 @@ django.jQuery(function ($) {
         $indoor.show();
         $indoorRows.show();
     }
+
+    var geocode_api = window.location.origin + '/admin/django_loci/location/geocode/';
+    var reverse_api = window.location.origin + '/admin/django_loci/location/reverse-geocode/';
+    window.setTimeout(function () {
+        window.address_lat = null;
+        window.address_lng = null;
```
nemesisdesign: in javascript, use camelCase notation for variables and function names, so `addressLng`, `receivedAlert` and so on, please change all variables accordingly

django_loci/static/django-loci/js/loci.js
```diff
@@ -377,4 +464,90 @@ django.jQuery(function ($) {
         $indoor.show();
         $indoorRows.show();
     }
+
+    var geocode_api = window.location.origin + '/admin/django_loci/location/geocode/';
```
nemesisdesign: here

django_loci/static/django-loci/js/loci.js
```diff
@@ -343,6 +344,92 @@ django.jQuery(function ($) {
...
+            if (map._layers.hasOwnProperty(layer)) {
+                if (map._layers[layer].hasOwnProperty('_layers')) {
+                    return map._layers[layer];
+                }
+            }
+        }
+    }
+
+    // set marker
+    function update_map(geocode_api) {
```
nemesisdesign: updateMap and all the others as well

django_loci/static/django-loci/js/loci.js
```diff
@@ -343,6 +344,92 @@ django.jQuery(function ($) {
...
+                                        if (window.confirm('Do you want to update map?')) {
+                                            get_featureGroup().removeLayer(get_marker());
+                                            get_featureGroup().addLayer(L.marker([data.lat, data.lng]));
+                                            $('.django-leaflet-raw-textarea').val(wkt);
+                                            window.address_lat = data.lat;
+                                            window.address_lng = data.lng;
+                                            window.cancled = false;
+                                            getMap().setView([data.lat, data.lng], 14);
+                                        } else {
+                                            window.cancled = true;
```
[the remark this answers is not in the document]
oltarzewskik: sorry typo

oltarzewskik:
@nemesisdesign I did everything as requested.
If user go to change view and add marker using "draw" tool address field will be update only when hitting save. (It was done before)
I added my own idea: if user add marker using "draw" tool, and whithout saving to db will try to edit marker then address will be updated when saving

django_loci/static/django-loci/js/loci.js
```diff
@@ -377,4 +497,120 @@ django.jQuery(function ($) {
...
+        $addressInput.keyup(function () {
+            window.clearTimeout(timeout);
+            timeout = window.setTimeout(function () {
+                if ($addressInput.val().length >= 10) {
+                    updateMap(window.geocodeApi);
+                }
+            }, 1000);
+        });
+
+        window.setInterval(function () {
```
nemesisdesign: can you explain what this code is doing?
It looks like you're periodically checking if the marker has coordinates, but I think we can listen to leaflet events instead of doing it this way.

I have tested this and doesn't really work as expected: it doesn't ask me for confirmation to update the address if the address field is already filled.
Also if I delete the marker I noticed a JS error that breaks this feature completely:
```
loci.js:443 Uncaught TypeError: Cannot read property 'dragging' of undefined
    at updateAdress (loci.js:443)
```

django_loci/static/django-loci/js/loci.js
```diff
@@ -377,4 +497,120 @@ django.jQuery(function ($) {
...
+                var reverseApiWithData = window.reverseApi + '?lat=' + currentMarkerLat + '&lng=' + currentMarkerLng;
+                window.fetch(reverseApiWithData)
+                    .then(function (resp) {
+                        var dataPromise = resp.json();
+                        dataPromise.then(function (data) {
+                            if ($addressInput.val()) {
+                                if (currentMarkerLat !== window.addressLat ||
+                                        currentMarkerLng !== window.addressLng) {
+                                    if (resp.status === 200) {
+                                        if (window.confirm('Would you like to update address?')) {
```
nemesisdesign: I think we should ask the user for confirmation before doing the HTTP request, otherwise it looks like kinda slow.
Best thing would be to do the request in parallel to asking for confirmation, that way it would result in very smooth UI.

django_loci/static/django-loci/js/loci.js
oltarzewskik: > I have tested this and doesn't really work as expected: it doesn't ask me for confirmation to update the address if the address field is already filled.

It asks you after hitting save (django save), but I am not sure what do you mean. Could you give me step by step how to reproduce?

> Also if I delete the marker I noticed a JS error that breaks this feature completely:

same here

TheOneAboveAllTitan:
@oltarzewskik feel free to ask for help. :smile:

oltarzewskik:
@TheOneAboveAllTitan I am not good at js so I am not sure what @nemesisdesign, but it looks like I have to do some async code. If I will find any difficulty I will ping :+1:

django_loci/static/django-loci/js/loci.js
oltarzewskik: so behavior should be following:
Ask user for confirmation and while waiting for user's confirmation asynchronously do request to API, and then after affirmative answer to popup, instantly update address field?

nemesisdesign: that would be awesome!

django_loci/static/django-loci/js/loci.js
nemesisdesign: not after django save.. after the widget of django-leaflet save, as explained in this comment: https://github.com/openwisp/django-loci/pull/46#pullrequestreview-347817033

oltarzewskik: > Also if I delete the marker I noticed a JS error that breaks this feature completely:
loci.js:443 Uncaught TypeError: Cannot read property 'dragging' of undefined
    at updateAdress (loci.js:443)

I am not able to reproduce this part anyway, make sure that code is rebased

oltarzewskik:
Todo list:

- [ ] Move to async code as explained here: https://github.com/openwisp/django-loci/pull/46#discussion_r386015710

- [ ] Make popup which will ask if user wants to update address, after moving marker via "Edit layers"

- [ ]  Try to listen to leaflet events instead of periodically checking if the marker has coordinates

django_loci/static/django-loci/js/loci.js
```diff
@@ -377,4 +497,120 @@ django.jQuery(function ($) {
...
+                                    submit = true;
+                                    $form.submit();
+                                }
+                            } else {
+                                $addressInput.val(data.address);
+                                submit = true;
+                                $form.submit();
+                            }
+                        });
+                    });
```
an0Kx42:
```suggestion
        $form.submit(async function (e) {
            var currentMarkerCoords = getMarker() !== undefined ? getMarker().getLatLng() : null;
            var currentMarkerLat = getMarker() !== undefined ? getMarker().getLatLng().lat : null;
            var currentMarkerLng = getMarker() !== undefined ? getMarker().getLatLng().lng : null;
            if (currentMarkerCoords !== window.markerCoords && currentMarkerCoords !== null &&
                    getMarker().dragging._enabled !== true) {
                var reverseApiWithData = window.reverseApi + '?lat=' + currentMarkerLat + '&lng=' + currentMarkerLng;
                var resp = window.fetch(reverseApiWithData)
                if (!window.confirm('Would you like to update address?')) {
                    submit = true;
                    $form.submit();
                    return;
                }
                resp = await resp;
                var data = await resp.json();
                if ($addressInput.val()) {
                	if (currentMarkerLat !== window.addressLat ||
                	currentMarkerLng !== window.addressLng) {
                        if (resp.status === 200) {
                            $addressInput.val(data.address);
                            submit = true;
                            $form.submit();
                        } else {
                            var message = 'We couldn\'t find a related address to the location '
                                           + 'selected on the map, please enter the address manually';
                            alert(message);
                            window.skipGeocode = 2;
                        }
                    } else {
                        submit = true;
                        $form.submit();
                    }
                } else {
                    $addressInput.val(data.address);
                    submit = true;
                    $form.submit();
                }
```
Asynchronously do request to API, while waiting for user's confirmation.

oltarzewskik: It doesn't work, but anyway thanks for making base for this feature :smile:

django_loci/static/django-loci/js/loci.js
oltarzewskik: Done :+1:

PR #51, opened 2020-01-19
[chores] Removed django<1.11 code

oltarzewskik:


django_loci/channels/base.py
```diff
@@ -42,8 +42,6 @@ def connect(self):
     def is_authorized(self, user, location):
         perm = '{0}.change_location'.format(self.model._meta.app_label)
         authenticated = user.is_authenticated
-        if callable(authenticated):  # pragma: nocover
```
atb00ker: Great. It's not required indeed.
However, it's not an `[enchancement]` -> `[chores]` is a better word! :smile:
:+1:

PR #55, opened 2020-03-05
[geo] Fixed JS logic crashes #c35

### LINKED ISSUE #35: [ux] Floorplan fields is shown when error occurs in new indoor environment
To produce this, follow these steps:
1. Add device, leave some fields empty (especially the required one: name, organization and mac address) to produce error when saving.
2. In `Geographic Information` secton, set `Location selection` to **New** and set Type to **Indoor environment**
3. In `Indoor Coordinates` section, select file to the Image field, and set the position for Indoor (you can just use dummy image and set position anywhere) to bypass the client indoor position validation.
4. Save the device.
5. In `Indoor Coordinates` section, you can see the `Floorplan section` and `Floorplan` fields is showed.

The expected output: Floorplan fields should be hidden (You can change the type to Outdoor and change it again to Indoor, then you see Floorplan fields is hidden again).
Additional note: The Floorplan fields should be showed **only in existing location**

NoumbissiValere:
The issue was due to the fact that the geometry
is set to None when is_mobile is True and we were
try to get the json of NoneType.
I resolved the problem by making sure the json
function is not called on a NoneType
This fixes [this issue](https://github.com/openwisp/openwisp-controller/issues/35) in openwisp_controller

Fixes #c35

### DIFF
--- a/django_loci/base/admin.py
+++ b/django_loci/base/admin.py
@@ -78,7 +78,7 @@ class AbstractLocationAdmin(TimeReadonlyAdminMixin, LeafletGeoAdmin):
             'type': instance.type,
             'is_mobile': instance.is_mobile,
             'address': instance.address,
-            'geometry': json.loads(instance.geometry.json)
+            'geometry': json.loads(instance.geometry.json) if instance.geometry else None
         })
 
     def floorplans_json_view(self, request, pk):
--- a/django_loci/static/django-loci/js/loci.js
+++ b/django_loci/static/django-loci/js/loci.js
@@ -134,7 +134,10 @@ django.jQuery(function ($) {
         ];
         if ($isMobile.prop('checked')) {
             $(rows).each(function (i, el) {
-                $(el).hide();
+                if (($(el).attr('class').includes('address') && $address.val()) ||
+                        ($(el).attr('class').includes('geometry') && $geometryTextarea.val())) {
+                    $(el).show();
+                } else { $(el).hide(); }
             });
         } else {
             $(rows).each(function (i, el) {
@@ -207,7 +210,10 @@ django.jQuery(function ($) {
     function locationChange(e, initial) {
         function loadIndoor() {
             indoorForm();
-            if ($type.val() !== 'indoor') { return; }
+            if ($type.val() !== 'indoor') {
+                $indoor.hide();
+                return;
+            }
             var floorplansUrl = getLocationFloorplansJsonUrl($location.val());
             $.getJSON(floorplansUrl, function (data) {
                 var $current = $floorplan.find('option:selected'),
@@ -237,11 +243,12 @@ django.jQuery(function ($) {
                 $type.val(data.type);
                 $isMobile.prop('checked', data.is_mobile);
                 $address.val(data.address);
-                $geometryTextarea.val(JSON.stringify(data.geometry));
+                $geometryTextarea.val(data.geometry ? JSON.stringify(data.geometry) : '');
                 var map = getMap();
                 if (map) { map.remove(); }
                 $geoEdit.show();
                 window[loadMapName]();
+                isMobileChange();
                 loadIndoor();
             });
         } else {
@@ -358,7 +365,7 @@ django.jQuery(function ($) {
         $locationSelection.parents('.form-row').hide();
         $locationRow.hide();
         $name.parents('.form-row').hide();
-        $address.parents('.form-row').hide();
+        if (!$address.val()) { $address.parents('.form-row').hide(); }
         // if no location data yet
         if (!$geometryTextarea.val()) {
             $geometryRow.hide();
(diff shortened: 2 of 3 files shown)


NoumbissiValere:
@nemesisdesign The pull request is ready. Please can you review it and give me feedback? 
Thanks :smile:

django_loci/tests/base/test_admin.py
```diff
@@ -66,3 +66,9 @@ def test_floorplan_change_image_removed(self):
         url = reverse('{0}_floorplan_change'.format(self.url_prefix), args=[fl.pk])
         r = self.client.get(url)
         self.assertContains(r, 'test-admin-location-1')
+
+    def test_is_mobile_location_json_view(self):
+        self._login_as_admin()
+        loc = self._create_location(is_mobile=True, geometry=None)
+        response = self.client.get(reverse('admin:django_loci_location_json', args=[loc.pk]))
+        self.assertEqual(response.status_code, 200)
```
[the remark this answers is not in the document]
NoumbissiValere: @nemesisdesign please what do you mean by the relevant part of the response? Is it checking if the content of our response is actually what we were expecting

django_loci/base/admin.py
```diff
@@ -73,13 +73,17 @@ def get_urls(self):
 
     def json_view(self, request, pk):
         instance = get_object_or_404(self.model, pk=pk)
-        return JsonResponse({
+        data = {
             'name': instance.name,
             'type': instance.type,
             'is_mobile': instance.is_mobile,
-            'address': instance.address,
-            'geometry': json.loads(instance.geometry.json)
```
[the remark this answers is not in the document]
NoumbissiValere: @nemesisdesign I tried setting geometry to `None` and `""` but each of them brought some `js` errors at the javascript console. but eliminating the geometry didn't produce any error and the UI worked well reason why i adopt this one. Please I don't know if you have any other suggestion to a safer way. 
Is there a link where i can study the different types of geometry that can be used, I tried finding that in the code, but found none, maybe i can find a safer way by looking at the various format of geometry which can be used.
Thanks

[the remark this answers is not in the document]
nemesisdesign: which line of JS is failing? I think we need to adapt the frontend anyway.

Removing properties from API response is not a best practice, it's obscure and confusing.
If you're working with the API and you get a different response you may not understand why. If you get a null geometry you can understand it instead I believe.

django_loci/tests/base/test_admin.py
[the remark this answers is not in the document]
nemesisdesign: checking the geometry property (either present or None).

NoumbissiValere:
Ok @nemesisdesign thanks for the feedback, I will work on it

NoumbissiValere:
@nemesisdesign I have made the updates

NoumbissiValere:
@nemesisdesign This is ready for review. Am very sorry for the delay.

django_loci/tests/base/test_admin.py
```diff
@@ -66,3 +66,27 @@ def test_floorplan_change_image_removed(self):
...
+        expected = {
+            'name': 'location2',
+            'address': 'loc2 add',
+            'type': 'outdoor',
+            'is_mobile': False,
+            'geometry': {
+                'type': 'Point',
+                'coordinates': [12.512124, 41.898903]
+            }
+
```
nemesisdesign: remove this blank line

django_loci/static/django-loci/js/loci.js
```diff
@@ -237,11 +240,14 @@ django.jQuery(function ($) {
...
                 $address.val(data.address);
-                $geometryTextarea.val(JSON.stringify(data.geometry));
+                $geometryTextarea.val(data.geometry ? JSON.stringify(data.geometry) : '');
                 var map = getMap();
                 if (map) { map.remove(); }
                 $geoEdit.show();
                 window[loadMapName]();
+                isMobileChange();
+                if ($address.val()) { $address.parents('.form-row').show(); }
+                if ($geometryTextarea.val()) { $geometryRow.show(); }
```
nemesisdesign: I think it makes sens to move these checks in the `isMobileChange` function

django_loci/static/django-loci/js/loci.js
```diff
@@ -207,7 +207,10 @@ django.jQuery(function ($) {
     function locationChange(e, initial) {
         function loadIndoor() {
             indoorForm();
-            if ($type.val() !== 'indoor') { return; }
+            if ($type.val() !== 'indoor') {
+                $indoor.hide();
```
nemesisdesign: why is this necessary?

NoumbissiValere: While i was testing the work i did, I realized that if i selected a location of type `indoor`, the `indoor fields` showed up which was very good, but when i changed the type back to `outdoor`, i expected the `indoor fields` to be hidden again but that was not the case, reason while i added the `$indoor.hide()` when `type != 'indoor'` 
no
Actually it had no connection with the issue i was solving but i noticed it while i was testing and taught of doing of that.

nemesisdesign: ok, I will double check this. Please follow up on the rest of the comments when you can, hopefully we'll be done with this soon :blush:

NoumbissiValere:
@nemesisdesign Please this is ready for review

PR #58, opened 2020-03-20
[model] Allow indoor location without indoor coords #54

### LINKED ISSUE #54: [model] Allow indoor location without indoor coords
When adding a location using the base class `AbstractObjectLocationInline`, we may want to flag that the object is indoor, but we may not have the floorplan image ready yet, in that case we are forced to set it to outdoor in order to save, but it's incorrect since it's not an outdoor location.

We could ask for confirmation to the user that they want to save the object even if they have not specified the indoor coordinates and if they confirm we can save the object anyway.

Changes to the JS and admin form will be necessary.

NoumbissiValere:
Fixed the issue by adding a confirmation script to the
loci.js which gives the opportunity to the admin to
confirm if he/she wish to save the indoor location
without indoor coords. Also modified the admin to make
sure this is saved without validation errors.
I also modified the model of object_location to allow
indoors to be None when type is indoor and no floorplan
is provided.

Fixes #54

### DIFF
--- a/django_loci/static/django-loci/js/loci.js
+++ b/django_loci/static/django-loci/js/loci.js
@@ -1,3 +1,6 @@
+/*global
+alert, confirm, console, Debug, opera, prompt, WSH
+*/
 /*
 this JS is shared between:
     - DeviceLocationForm
@@ -327,8 +330,14 @@ django.jQuery(function ($) {
     $('#content-main form').submit(function (e) {
         var indoorPosition = $('.field-indoor .floorplan-raw input').val();
         if ($type.val() === 'indoor' && !indoorPosition) {
-            e.preventDefault();
-            alert(gettext('Please set the indoor position before saving'));
+            if ($floorplanImage[0].files.length !== 0 && !indoorPosition) {
+                alert('Please set the indoor position before saving');
+                e.preventDefault();
+            } else if ($floorplanImage[0].files.length === 0 && !indoorPosition) {
+                var confirmation = confirm('Do you want to save this indoor' +
+                                        'location without indoor coordinates?');
+                if (!confirmation) { e.preventDefault(); }
+            }
         }
     });
 
--- a/django_loci/tests/base/test_models.py
+++ b/django_loci/tests/base/test_models.py
@@ -124,12 +124,15 @@ class BaseTestModels(TestLociMixin):
         self._test_indoor_position_validation_error(ol)
         ol.indoor = 'TOTALLY.WRONG'
         self._test_indoor_position_validation_error(ol)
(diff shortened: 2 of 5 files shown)


NoumbissiValere:
@nemesisdesign Please can you review this? 
Thanks

django_loci/static/django-loci/js/loci.js
```diff
@@ -327,8 +330,9 @@ django.jQuery(function ($) {
     $('#content-main form').submit(function (e) {
         var indoorPosition = $('.field-indoor .floorplan-raw input').val();
         if ($type.val() === 'indoor' && !indoorPosition) {
-            e.preventDefault();
-            alert(gettext('Please set the indoor position before saving'));
+            var confirmation = confirm('Do you want to save this indoor' +
+                                    'location without indoor coordinates?');
+            if (!confirmation) { e.preventDefault(); }
```
nemesisdesign: we should allow to save only if no floorplan is available.

If the floorplan is availabe, we must still force the user to specify the coordinates.

django_loci/tests/base/test_models.py
```diff
@@ -124,12 +124,15 @@ def test_invalid_indoor_position(self):
         self._test_indoor_position_validation_error(ol)
         ol.indoor = 'TOTALLY.WRONG'
         self._test_indoor_position_validation_error(ol)
+        ol.indoor = '100.2300,-45.23454'
+        ol.full_clean()
+        # allow empty indoor for location whose indoor coordinates are not yet received
         ol.indoor = ''
```
nemesisdesign: as said above, only allow this if no floorplan is available

NoumbissiValere:
Ok thanks for reviewing. I will review the work and make these points are met

django_loci/static/django-loci/js/loci.js
```diff
@@ -327,8 +330,14 @@ django.jQuery(function ($) {
     $('#content-main form').submit(function (e) {
         var indoorPosition = $('.field-indoor .floorplan-raw input').val();
         if ($type.val() === 'indoor' && !indoorPosition) {
-            e.preventDefault();
-            alert(gettext('Please set the indoor position before saving'));
+            if ($floorplanImage[0].files.length !== 0 && !indoorPosition) {
+                alert('Please set the indoor position before saving');
```
nemesisdesign, line 334: you should have used `gettext`

NoumbissiValere: Thank you for pointing this out

NoumbissiValere:
Happy this work finally got in :smile:

PR #64, opened 2020-07-07
[bug] Websocket connect error #63

### LINKED ISSUE #63: [bug] websocket connect error in location change form page
1. Create a location, set the marker and then check `is_mobile`.
2. Save the location.
3. Open it again and look at the console. (`Firefox can’t establish a connection to the server at ws://127.0.0.1:8000/ws/loci/location/undefined/.`)

atb00ker:
Closes #63

### DIFF
--- a/django_loci/static/django-loci/js/loci.js
+++ b/django_loci/static/django-loci/js/loci.js
@@ -147,7 +147,7 @@ django.jQuery(function ($) {
             });
             // name not relevant in mobile locations
             $name.parents('.form-row').hide();
-            $('.no-location').show();
+            if (!$geometryTextarea.val()) { $('.no-location').show(); }
         } else {
             $(rows).each(function (i, el) {
                 $(el).parents('.form-row').show();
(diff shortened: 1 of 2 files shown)


django_loci/static/django-loci/js/loci.js
```diff
@@ -147,7 +147,7 @@ django.jQuery(function ($) {
             });
             // name not relevant in mobile locations
             $name.parents('.form-row').hide();
-            $('.no-location').show();
+            if (!$geometryTextarea.val()) { $('.no-location').show(); }
```
atb00ker, line 150: If `$geometryTextarea.val()` has a value, it's not hidden. Hence, map and the message appear together, this condition will help avoid that.

django_loci/static/django-loci/js/loci.js
```diff
@@ -371,14 +371,18 @@ django.jQuery(function ($) {
     }
 
     // show existing location
+    var pk;
     if ($location.val()) {
         $locationSelectionRow.hide();
         $geoEdit.show();
         isNew = false;
+        pk = $location.val();
```
atb00ker, line 379: Changing `pk` value here should do the trick.

PR #67, opened 2020-07-25
0.3.2 release

atb00ker:


CHANGES.rst
```diff
@@ -1,6 +1,17 @@
...
+Version 0.3.3 [2020-07-25]
+--------------------------
+
+- [fix] Fixed websocket connect error for location change view
+- [deps] Added support for openwisp-utils~=0.5.1 and dropped previous versions
+- [deps] Added support for pillow 7.2.0 and dropped lower versions
+- [deps] Added support for pytest-django 3.9
+- [deps] Added support for pytest-asyncio 0.14
+- [deps] Added support for pytest-cov 0.10
+- [deps] Added support for django-leaflet 0.28
```
nemesisdesign, line 13: the test dependencies are not useful (users don't care, we just add noise in their eyes, the other dependencies can be summarized on one line, I'd put them in this order: pillow, django-leaflet and openwisp-utils

atb00ker:
> @atb00ker have you tested the admin to make sure the new version of django-leaflet does not cause issues?

Yes, it works, except when I drag the map with mouse it show error in `leftlet.js` on console but it does that for older versions as well and everything works even with the error, so I think we are in the green.

PR #72, opened 2020-10-13
[feature] Auto calculate geo coords or addr

### LINKED ISSUE #44: [ux] Automatically calculate geographic coordinates or address

NoumbissiValere:
Closes #44
Closes #46

### DIFF
--- a/django_loci/static/django-loci/js/loci.js
+++ b/django_loci/static/django-loci/js/loci.js
@@ -430,7 +430,8 @@ django.jQuery(function ($) {
                 } else {
                     var latlng = marker.getLatLng();
                     if (latlng.lat !== data.lat || latlng.lng !== data.lng) {
-                        message = gettext('Are you sure you wish to change the location?');
+                        message = gettext('The address was changed, would you like to ' +
+                                          'automatically update the location on the map?');
                         if (confirm(message)) {
                             updateLatLng(data);
                             featureGroup.removeLayer(marker);
@@ -448,11 +449,12 @@ django.jQuery(function ($) {
 
     function updateAdress() {
         var marker = getMarker(),
-            message;
+            message,
+            latlng;
         if (marker === undefined) {
             return;
         }
-        var latlng = marker.getLatLng();
+        latlng = marker.getLatLng();
         if (latlng.lat.toString() === $oldLat && latlng.lng.toString() === $oldLng) {
             return;
         }
@@ -462,22 +464,25 @@ django.jQuery(function ($) {
                 if (!$addressInput.val()) {
                     $addressInput.val(data.address);
                 } else {
-                    message = gettext('Do you wish to change this address?');
+                    message = gettext('The location on the map was changed, would you ' +
+                                      'like to update the address to');
+                    message += ' "' + data.address + '"?';
                     if (confirm(message)) {
                         $addressInput.val(data.address);
                     }
                 }
             })
             .fail(function () {
-                message = gettext('Could not find address related to the location.');
+                message = gettext('Could not find any address related to this location.');
                 alert(message);
             });
     }
 
-    // drag marker event
-    function updateAddrOnDrag() {
+    // triggers update of the address when the location on the map is changed
+    function updateAddressOnMapChange() {
         var marker = getMarker();
-        marker.on('dragend', function () {
+        if (!marker) { return; }
+        getMap().on('draw:edited', function (e) {
             updateAdress();
             updateMapView(marker.getLatLng());
         });
@@ -492,13 +497,14 @@ django.jQuery(function ($) {
             marker = getMarker();
         featureGroup.on('layeradd', function () {
             updateAdress();
-            updateAddrOnDrag();
+            updateAddressOnMapChange();
             marker = getMarker();
+            if (!marker) { return; }
             updateMapView(marker.getLatLng());
         });
         if (marker !== undefined) {
             updateLatLng(marker.getLatLng());
-            updateAddrOnDrag();
+            updateAddressOnMapChange();
         }
     });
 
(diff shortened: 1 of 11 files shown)


requirements.txt
```diff
@@ -2,5 +2,5 @@ django>=2.2,<3.1
 django-leaflet>=0.23,<0.27
 channels>=2.1.3,<2.5.0
 Pillow>=6.2.0,<=7.1.0
-openwisp-utils>=0.4.0,<0.5.0
```
NoumbissiValere: The error shown by `TravisBuddy` was due to the fact that `openwisp-utils version 0.4.0` was being  installed which doesn't have `openwisp-qa-checks`. 
This is the reason why i had to update it

NoumbissiValere:
@nemesisdesign I think the JavaScript work is ok now

nemesisdesign (changes requested):

requirements.txt
```diff
@@ -2,5 +2,5 @@ django>=2.2,<3.1
...
+openwisp-utils~=0.6.0
 geopy>=1.20.0,<1.22.0
```
nemesisdesign, line 6: geopy current latest version is 2.0, can you make sure we use that and the code works please?

nemesisdesign (commented):
Thanks @NoumbissiValere I'm testing it.

In the meanwhile, can you please write a test for these 3 lines to avoid the code coverage to go down? https://coveralls.io/builds/34244754/source?filename=django_loci%2Fapps.py#L33
You could instantiate the appconfig class and run that test singularly or call the ready() method, I think we do something similar in other modules, maybe openwisp-monitoring.

requirements.txt
```diff
@@ -1,5 +1,6 @@
-django>=2.2,<3.2
-django-leaflet>=0.23,<0.28
-channels>=2.3.0,<2.5.0
-Pillow~=7.2.0
+django>=2.2,<3.1
+django-leaflet>=0.23,<0.27
+channels>=2.1.3,<2.5.0
+Pillow>=6.2.0,<=7.1.0
```
nemesisdesign, line 4: Could you rebase on master so to include this change?
https://github.com/openwisp/django-loci/commit/a8cfe17a8c7d6291dc02f2192fe979cfbd0b13d7

django_loci/static/django-loci/js/loci.js
```diff
@@ -462,22 +464,25 @@ django.jQuery(function ($) {
...
     }
 
-    // drag marker event
-    function updateAddrOnDrag() {
+    // triggers update of the address when the location on the map is changed
+    function updateAddressOnMapChange() {
         var marker = getMarker();
-        marker.on('dragend', function () {
+        if (!marker) { return; }
+        getMap().on('draw:edited', function (e) {
```
nemesisdesign, line 485: @NoumbissiValere @oltarzewskik :eyes: :point_up: https://leaflet.github.io/Leaflet.draw/docs/leaflet-draw-latest.html#l-draw-event

nemesisdesign (approved):
Added some further improvements, now is ready, I suggest to test it again on your side to double check.

django_loci/static/django-loci/js/loci.js
NoumbissiValere: Nice, `edited` is cool. it fires up the callback only the save option is clicked which is perfect. thank you @nemesisdesign 
:+1:

NoumbissiValere:
@nemesisdesign I have tested this work again and it works better. I think it is ready.

nemesisdesign (approved):
Thanks @NoumbissiValere and thanks also @oltarzewskik for the initial work on this :+1:

PR #73, opened 2020-10-14
[bug] NOT NULL constraint failed: django_loci_floorplan.floor #65

### LINKED ISSUE #65: [bug] NOT NULL constraint failed: django_loci_floorplan.floor
```
IntegrityError at /admin/testdeviceapp/device/add/
NOT NULL constraint failed: django_loci_floorplan.floor
```

Steps to reproduce:
1. Create a device with new indoor environment, **flag it as mobile**
2. In the "Indoor coordinates" section, do not fill  a value for `floor`.
3. Complete the rest of the form correctly.
4. Submit

Purhan:
Closes #65

### DIFF
--- a/django_loci/base/admin.py
+++ b/django_loci/base/admin.py
@@ -246,7 +246,7 @@ class AbstractObjectLocationForm(forms.ModelForm):
         fields = []
         if not is_mobile and type_ in ['outdoor', 'indoor']:
             fields += ['location_selection', 'name', 'address', 'geometry']
-        if not is_mobile and type_ == 'indoor':
+        if type_ == 'indoor':
             if data.get('floorplan_selection') == 'existing':
                 fields.append('floorplan')
             if data.get('image'):
@@ -278,7 +278,7 @@ class AbstractObjectLocationForm(forms.ModelForm):
         floorplan = data.get('floorplan') or self.floorplan_model()
         floorplan.location = instance.location
         floor = data.get('floor')
-        floorplan.floor = floor if floor is not None else floorplan.floor
+        floorplan.floor = floor
         # the image path is updated only during creation
         # or if the image has been actually changed
         if data.get('image') and self.initial.get('image') != data.get('image'):
--- a/django_loci/base/admin.py
+++ b/django_loci/base/admin.py
@@ -277,8 +277,7 @@ class AbstractObjectLocationForm(forms.ModelForm):
         instance = self.instance
         floorplan = data.get('floorplan') or self.floorplan_model()
         floorplan.location = instance.location
-        floor = data.get('floor')
-        floorplan.floor = floor
+        floorplan.floor = data.get('floor')
         # the image path is updated only during creation
         # or if the image has been actually changed
         if data.get('image') and self.initial.get('image') != data.get('image'):
--- a/django_loci/tests/base/test_admin_inline.py
+++ b/django_loci/tests/base/test_admin_inline.py
@@ -672,6 +672,32 @@ class BaseTestAdminInline(TestAdminMixin, TestLociMixin):
         self.assertEqual(ol.location.type, 'indoor')
         self.assertEqual(ol.indoor, '')
 
+    def test_add_indoor_mobile_location_without_floor(self):
+        self._login_as_admin()
+        p = self._get_prefix()
+        floorplan_file = open(self._floorplan_path, 'rb')
+        params = self.params
+        params.update(
+            {
+                'name': 'test-add-indoor-mobile-location-without-floor',
+                '{0}-0-type'.format(p): 'indoor',
+                '{0}-0-location_selection'.format(p): 'new',
+                '{0}-0-is_mobile'.format(p): True,
+                '{0}-0-location'.format(p): '',
+                '{0}-0-floorplan_selection'.format(p): 'new',
+                '{0}-0-floorplan'.format(p): '',
+                '{0}-0-floor'.format(p): '',
+                '{0}-0-image'.format(p): floorplan_file,
+                '{0}-0-indoor'.format(p): '',
+                '{0}-0-id'.format(p): '',
+            }
+        )
+        r = self.client.post(reverse(self.add_url), params, follow=True)
+        floorplan_file.close()
+        self.assertContains(r, 'errors field-floor')
+        loc = self.location_model.objects.filter(name=params['{0}-0-name'.format(p)])
+        self.assertEqual(loc.count(), 0)
+
     def test_add_indoor_location_without_coords(self):
         self._login_as_admin()
         p = self._get_prefix()


Purhan:
@atb00ker please take a look

django_loci/base/models.py
```diff
@@ -85,7 +85,7 @@ def short_type(self):
 
 class AbstractFloorPlan(TimeStampedEditableModel):
     location = models.ForeignKey('django_loci.Location', on_delete=models.CASCADE)
-    floor = models.SmallIntegerField(_('floor'))
+    floor = models.SmallIntegerField(_('floor'), null=True)
```
atb00ker, line 88: Well, it works alright, but we should add a condition that it can't be null. (`null-false, blank=false`), because a floor always has number, right?

What do you think?

atb00ker (commented):
@Purhan, review below, try `blank=false` ! :smile:

django_loci/base/models.py
Purhan: Yes I figured. Should I add a default value as well?

nemesisdesign (changes requested):
We shouldn't save empty floorplan records.
We should make sure the floorplan records does not get saved, without errors.

django_loci/base/models.py
```diff
@@ -85,7 +85,7 @@ def short_type(self):
 
 class AbstractFloorPlan(TimeStampedEditableModel):
     location = models.ForeignKey('django_loci.Location', on_delete=models.CASCADE)
-    floor = models.SmallIntegerField(_('floor'))
+    floor = models.SmallIntegerField(_('floor'), null=False, blank=False)
```
nemesisdesign, line 88: this should not be necessary because these are the default values for fields in django, see:

- https://docs.djangoproject.com/en/3.1/ref/models/fields/#null
- https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.Field.blank

nemesisdesign (changes requested):

nemesisdesign (commented):
PS: looks like you're on the right path :wink:

django_loci/base/models.py
Purhan: Oh, I see. Thanks for letting me know! Going to remove that next time :smile:

Purhan:
@nemesisdesign I have a question. I am going to make it such that it does not let you save if type is "indoor" and floor is NULL. However, there are tests that do exactly that (i.e `test_add_indoor_location_without_indoor_coords`). What should I do in this case?

nemesisdesign (commented):
When fixing bugs, the first thing to do is to confirm the bug and write a failing test, I think there's a problem with this issue. Please see my comment on the issue here: https://github.com/openwisp/django-loci/issues/65#issuecomment-709477481

Keep this in mind for future bug fixes.

Purhan:
> When fixing bugs, the first thing to do is to confirm the bug and write a failing test, I think there's a problem with this issue. Please see my comment on the issue here: [#65 (comment)](https://github.com/openwisp/django-loci/issues/65#issuecomment-709477481)
> 
> Keep this in mind for future bug fixes.

Got it. Since I was able to replicate the issue and it was not covered by tests already, I have to write a test that fails particularly for a scenario like this and think of a suitable patch accordingly. Thanks.

Purhan:
@nemesisdesign @atb00ker 
As discussed in https://github.com/openwisp/django-loci/issues/65#issuecomment-709477481, we should allow users to save `null` as a value for floor. 
> because sometimes people need to create objects, flag them as indoor but they can't set the indoor coordinates in that moment, they defer that for the future

I have written tests that would fail for the steps mentioned in #65 and then went ahead to fix the errors thrown by the test. Please take a look.

atb00ker (commented):
It fixes the issue.

However, I noticed if floor number is set and then I edit the form to remove the floor number and save, the change is not honored and the hold floor number is retained.

Purhan:
> It fixes the issue.
> 
> However, I noticed if floor number is set and then I edit the form to remove the floor number and save, the change is not honored and the hold floor number is retained.

@atb00ker 
That should be fixed in the latest commit now (Waiting for checks to pass).

nemesisdesign (changes requested):
@Purhan please see what I wrote here: https://github.com/openwisp/django-loci/issues/65#issuecomment-710626150

For some reason the ValidationError is wrongly not being raised.
Looking at the code, it becomes evident, the floor field of the form is not marked as required: https://github.com/openwisp/django-loci/blob/9321eb1d225a9995d9c3702ff57a839971bbe8af/django_loci/base/admin.py#L155
Probably because we don't want to raise the validation error if the entire form is left empty because in that case we just won't save anything. But in this case, one field is supplied so the system thinks it should save.
The validation logic of the form must be changed to raise a validation error if any field is filled and floor is not filled.

Purhan:
@nemesisdesign, please check 174a620. Now a validation error is raised if and only if any other fields are filled.

django_loci/base/models.py
```diff
@@ -85,7 +85,7 @@ def short_type(self):
 
 class AbstractFloorPlan(TimeStampedEditableModel):
     location = models.ForeignKey('django_loci.Location', on_delete=models.CASCADE)
-    floor = models.SmallIntegerField(_('floor'))
+    floor = models.SmallIntegerField(_('floor'), null=True, blank=True)
```
nemesisdesign, line 88: this should not be changed

django_loci/base/admin.py
```diff
@@ -278,7 +281,7 @@ def _get_floorplan_instance(self):
         floorplan = data.get('floorplan') or self.floorplan_model()
         floorplan.location = instance.location
         floor = data.get('floor')
-        floorplan.floor = floor if floor is not None else floorplan.floor
+        floorplan.floor = floor
```
nemesisdesign, line 284: are you sure this should be changed?

django_loci/base/admin.py
```diff
@@ -256,6 +256,9 @@ def clean(self):
             data['address'] = ''
             data['geometry'] = ''
             data['location_selection'] = 'new'
+        if type_ == 'indoor' and is_mobile:
+            if (data.get('image') or data.get('location')) and ['floor'] not in fields:
+                fields += ['floor']
```
nemesisdesign, line 261: I think we could just change the existing code: 

https://github.com/openwisp/django-loci/blob/174a620e3e668095df90b47bcda4468af0a502ae/django_loci/base/admin.py#L249-L253

We could remove `not is_mobile and`, that may indeed be the cause of the bug.

nemesisdesign (changes requested):
Good progress @Purhan, see my comments below.

django_loci/base/admin.py
Purhan: As @atb00ker said above,

> However, I noticed if floor number is set and then I edit the form to remove the floor number and save, the change is not honored and the hold floor number is retained.

So I had to make this change in order to fix that.

nemesisdesign (approved):
Thanks a lot for your patience and good work @Purhan! :+1: 

I applied a minor improvement, see https://github.com/openwisp/django-loci/pull/73/commits/597bf1eb775fc56a67b78d750701871d5a157b06.

Will merge when the build finishes.

PR #75, opened 2020-10-22
[fix] Fix automatic geocoding in extended apps (eg: openwisp-controller) #44

### LINKED ISSUE #44: [ux] Automatically calculate geographic coordinates or address

NoumbissiValere:
- Fixed regression tests which were failing
on openwisp_controller because of poorly
setup url

- Made map's urls dynamic

Closes #44

### DIFF
--- a/django_loci/static/django-loci/js/loci.js
+++ b/django_loci/static/django-loci/js/loci.js
@@ -47,8 +47,8 @@ django.jQuery(function ($) {
         $mapGeojsonTextarea = $('.django-leaflet-raw-textarea'),
         $oldLat,
         $oldLng,
-        $getCoordsUrl = window.location.origin + '/admin/django_loci/location/geocode/',
-        $getAddrUrl = window.location.origin + '/admin/django_loci/location/reverse-geocode/';
+        $coordsUrl = $('#loci-geocode-url').attr('data-url'),
+        $addrUrl = $('#loci-reverse-geocode-url').attr('data-url');
 
     // define dummy gettext if django i18n is not enabled
     if (!gettext) { window.gettext = function (text) { return text; }; }
@@ -420,7 +420,7 @@ django.jQuery(function ($) {
             getFeatureGroup().clearLayers();
             return;
         }
-        $.get($getCoordsUrl, {address: addressValue})
+        $.get($coordsUrl, {address: addressValue})
             .done(function (data) {
                 var marker = getMarker(),
                     featureGroup = getFeatureGroup();
@@ -459,7 +459,7 @@ django.jQuery(function ($) {
             return;
         }
         updateLatLng(latlng);
-        $.get($getAddrUrl, {lat: latlng.lat, lng: latlng.lng})
+        $.get($addrUrl, {lat: latlng.lat, lng: latlng.lng})
             .done(function (data) {
                 if (!$addressInput.val()) {
                     $addressInput.val(data.address);
--- a/django_loci/base/admin.py
+++ b/django_loci/base/admin.py
@@ -59,6 +59,8 @@ class AbstractLocationAdmin(TimeReadonlyAdminMixin, LeafletGeoAdmin):
     search_fields = ['name', 'address']
     list_filter = ['type', 'is_mobile']
     save_on_top = True
+    
+    # This allows apps which extend django-loci to load this template with less hacks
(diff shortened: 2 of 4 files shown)


django_loci/static/django-loci/js/loci.js
```diff
@@ -508,6 +518,14 @@ django.jQuery(function ($) {
...
     });
 
+    var pathname = window.location.pathname;
+    if (pathname.includes('django_loci') || pathname.includes('testdeviceapp')) {
+        $coordsUrl = getCoordsUrl('django_loci');
+        $addrUrl = getAddrUrl('django_loci');
+    } else if (pathname.includes('geo') || pathname.includes('device')) {
+        $coordsUrl = getCoordsUrl('geo');
+        $addrUrl = getAddrUrl('geo');
+    }
```
nemesisdesign, line 528: Unfortunately this is not a good approach, because "geo" is from openwisp-controller, you're adding logic that knows details of openwisp-controller, violating the loose coupling principle.
It would also mean that any other app which extends this library and is not called "geo" will not work.

This issue was already addressed for the other URLs to get the floorplan data, there's an inline template override which generates the URL using the django view name, these are retrieved with javascript.
See:
- https://github.com/openwisp/django-loci/blob/master/django_loci/templates/admin/django_loci/location_inline.html
- https://github.com/openwisp/django-loci/blob/master/django_loci/static/django-loci/js/loci.js#L34-L35

You simply have to be consistent with what's already done, please keep this in mind for all subsequent pull requests.

nemesisdesign (changes requested):
Almost there, but there's one issue regarding loose coupling.
See below.

The commit type is not a chore but fix (see the title I renamed).

NoumbissiValere:
:smile: I knew it was a bad idea making `django_loci` aware of `geo`. It goes against the principle of orthogonality. Thanks @nemesisdesign I will fix this asap

nemesisdesign (approved):
It works! Thanks @NoumbissiValere :+1: :clap: 

I added a comment to the code, see 8f9ba43b455898fee5821ea4ea2fbaa25e57dd1f.

PR #76, opened 2020-10-24
[fix] Ignores is_mobile False from device #62

### LINKED ISSUE #62: [bug] ignores `is_mobile` False from device
In admin interface:
1. Create a device with a location, set the marker, check `is_mobile` & save.
2. Open it again, uncheck `is_mobile`, save
3. Open the device again and notice `is_mobile` is checked.

Purhan:
Fixes #62

### DIFF
--- a/django_loci/base/admin.py
+++ b/django_loci/base/admin.py
@@ -280,7 +280,7 @@ class AbstractObjectLocationForm(forms.ModelForm):
         data = self.cleaned_data
         location = data.get('location') or self.location_model()
         location.type = data.get('type') or location.type
-        location.is_mobile = data.get('is_mobile') or location.is_mobile
+        location.is_mobile = data.get('is_mobile', location.is_mobile)
         location.name = data.get('name') or location.name
         location.address = data.get('address') or location.address
         location.geometry = data.get('geometry') or location.geometry
--- a/django_loci/tests/base/test_admin_inline.py
+++ b/django_loci/tests/base/test_admin_inline.py
@@ -514,6 +514,35 @@ class BaseTestAdminInline(TestAdminMixin, TestLociMixin):
             loc.objectlocation_set.first().content_object.name, params['name']
         )
 
+    def test_remove_mobile(self):
+        self._login_as_admin()
+        p = self._get_prefix()
+        obj = self._create_object(name='test-remove-mobile')
+        pre_loc = self._create_location(name=obj.name, is_mobile=True)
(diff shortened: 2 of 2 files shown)


django_loci/base/admin.py
```diff
@@ -280,7 +280,7 @@ def _get_location_instance(self):
         data = self.cleaned_data
         location = data.get('location') or self.location_model()
         location.type = data.get('type') or location.type
-        location.is_mobile = data.get('is_mobile') or location.is_mobile
```
NoumbissiValere, line 283: @Purhan I don't support the idea of removing the `or` part of this statement because it must have been put there for some reason. this solution fixes the problem but it can cause issues in other modules which depends on this one. i think a good fix should be:
```location.is_mobile = location.is_mobile if data.get('is_mobile') is None else data.get('is_mobile')```

this fixes our problem and also maintain the original logic. 

What do you think?

**Review by NoumbissiValere — CHANGES_REQUESTED** [review]
Great work @Purhan  please see my comment below and also change `[bug]` in your commit message to `[fix]` and `Closes` to `Fixes` since you are providing a fix to the bug

File: django_loci/base/admin.py (hunk 1 above)
Comment (Purhan) [reply django_loci/base/admin.py:283]: Thank you, I was also skeptical about that change beforehand :smile:. I will quickly make these changes.

**Review by NoumbissiValere — DISMISSED** [review]
Great. This looks good to me :+1: . Leaving for @nemesisdesign to also test then merge

File: django_loci/base/admin.py [hunk 2]
```diff
@@ -280,7 +280,11 @@ def _get_location_instance(self):
...
+        location.is_mobile = (
+            location.is_mobile
+            if data.get('is_mobile') is None
+            else data.get('is_mobile')
+        )
```
atb00ker, line 287: Hate to be late here but there is a simpler solution:
Checkout this [best practice](http://docs.quantifiedcode.com/python-anti-patterns/correctness/not_using_get_to_return_a_default_value_from_a_dictionary.html).

We can do: `data.get('is_mobile', location.is_mobile)` for the same result.


Please test if this is practical in this particular place. (I haven't tested, just thought of using `default` in the `.get()` :smile: )

Purhan: Done. Works as expected. Thank you :smile:

atb00ker (approved):
Manually tested as well, LGTM! :-)

nemesisdesign (approved):
Thanks @Purhan @atb00ker :+1:

PR #80, opened 2021-02-16
[deps] Upgraded channels~=3.0.2 #79

### LINKED ISSUE #79: [deps] Update channels and pytest
[Channels 3.0 brought some breaking changes](https://channels.readthedocs.io/en/latest/releases/3.0.0.html).

This upgrade has to be coordinated with other OpenWISP modules (mainly notifications and controller) which also use django-channels.

niteshsinha17:
closes #79


niteshsinha17:
After completing this and https://github.com/openwisp/openwisp-notifications/pull/156 we can move on to update channels in controller

nemesisdesign (approved):
Looks good @niteshsinha17!

I think you can proceed with ow-controller.

What do you guys think? @pandafy @atb00ker

**PS**: I tested it (ran tests and tried running it in the browser, trying the mobile location feature and changing the mobile location to see if the map position is updated).

django_loci/channels/asgi.py
```diff
@@ -6,15 +6,15 @@
 from django_loci.channels.base import location_broadcast_path
 from django_loci.channels.consumers import LocationBroadcast
 
-channel_routing = ProtocolTypeRouter(
+application = ProtocolTypeRouter(
```
atb00ker, line 9: LGTM, except `application` is a very generic variable name, if any library in any of the dependencies use this name, we would have to change it in the future, what do you think?

niteshsinha17: I think its being used in this particular file only. We can continue with this name.

pandafy: Good catch Ajay! I overlooked this. It should stay `channel_routing` to make it consistent with code of [openwisp_controller/geo/channels/routing.py](https://github.com/openwisp/openwisp-controller/blob/b1ed757eded66101a1e1f74a5f3aa71e88fe7810/openwisp_controller/geo/channels/routing.py#L20-L26)

atb00ker: > I think its being used in this particular file only. We can continue with this name.

I understand but I generally fear names that are too general. `website`, `application`, `variable`, `flag`, `input` etc.
All are potential pointless dependency related bugs staring us at our faces.

niteshsinha17: > Good catch Ajay! I overlooked this. It should stay channel_routing to make it consistent with code of openwisp_controller/geo/channels/routing.py


Okay, then I will change it. I will need to change the name in this PR too https://github.com/openwisp/openwisp-notifications/pull/156.

atb00ker: Thanks @niteshsinha17

pandafy: >  I will need to change the name in this PR too openwisp/openwisp-notifications#156.

No! The following code in openwisp-notifications exists in `tests/openwisp2/routing.py`

```python
application = ProtocolTypeRouter(
        {'websocket': AuthMiddlewareStack(URLRouter(get_routes(consumers)))}
    )
```
Since it lives in the test project, it does not get shipped(and hence, does not matter much). 

The code here in question gets shipped though. So, it is wiser to avoid using such names that could raise conflicts in the future. 

I hope this explanation helps you. :smile:

niteshsinha17: I got your point. Thanks @pandafy

niteshsinha17:
Modified required files. Ready for review.

atb00ker (approved):
Tested, LGTM! :smile:
Let's upgrade channels everywhere simultaneous! :smile:

nemesisdesign (dismissed):
Cool, thank you @niteshsinha17!
Now with [notfications](https://github.com/openwisp/openwisp-notifications/pull/156) and loci done, we need to take care of [openwisp-controller](https://github.com/openwisp/openwisp-controller/issues/324).

PR #81, opened 2021-02-18
[fix] Fix strict_test setting naming

kkreitmair:
Setting name of `DJANGO_LOCI_GEOCODE_STRICT_TEST` should be the same as in settings, app_settings and docs.

### DIFF
--- a/django_loci/settings.py
+++ b/django_loci/settings.py
@@ -4,7 +4,7 @@ from django.utils.module_loading import import_string
 
 DJANGO_LOCI_GEOCODER = getattr(settings, 'DJANGO_LOCI_GEOCODER', 'ArcGIS')
 DJANGO_LOCI_GEOCODE_STRICT_TEST = getattr(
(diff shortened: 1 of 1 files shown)


okraits:
@kkreitmair @nemesisdesign What about adding a hint/note to the README that DNS and internet access is needed for the test to be successful?

In our case DNS and internet access was not available which led to gunicorn timeouts and strange behaviour which was quite hard to debug.

nemesisdesign (commented):
What a nasty bug! I didn't see this! Thank you so much.
I recently set this setting to false on an instance and noticed the same issue happening anyway but didn't have time to debug, so that's why it was happening.

@okraits if you think that is useful to have it, we can surely add a brief note here: https://github.com/openwisp/django-loci#django-loci-geocode-strict-test, go ahead and add it when you can

nemesisdesign (approved):

PR #84, opened 2021-03-05
[qa] Replace external services with mocks #83

### LINKED ISSUE #83: [bug] Geocoding operations should not hit external services during tests but should be mocked
Try to run the tests of django-loci while offline.

Result: some tests fail because they perform real geocoding operations.

We need to fix this, we should use mocking, I have used the [responses](https://github.com/getsentry/responses) library for similar simple HTTP request mocking operations successfully in the past and we could reuse it here if applicable.

The tests that have to be fixed are:

- ``test_geocode``
- ``test_reverse_geocode``

Once the response from the geocoding server is mocked, the real web service won't be called anymore during tests, therefore we expect to run tests while being offline with no failures.

This issue is affecting also openwisp-controller, because openwisp-controller imports the test suite from django-loci and executes it.

Purhan:
Closes #83


Purhan:
Although all tests pass after this, some tests still raise exceptions, example:
```
Traceback (most recent call last):
  File "/home/purhan/foss/django-loci/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/home/purhan/foss/django-loci/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 382, in _make_request
    self._validate_conn(conn)
  File "/home/purhan/foss/django-loci/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 1010, in _validate_conn
    conn.connect()
  File "/home/purhan/foss/django-loci/venv/lib/python3.8/site-packages/urllib3/connection.py", line 353, in connect
    conn = self._new_conn()
  File "/home/purhan/foss/django-loci/venv/lib/python3.8/site-packages/urllib3/connection.py", line 181, in _new_conn
    raise NewConnectionError(
urllib3.exceptions.NewConnectionError: <urllib3.connection.HTTPSConnection object at 0x7f28ddb774f0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution
```

requirements-test.txt
```diff
@@ -4,3 +4,4 @@ pytest-django>=3.4.2,<3.10.0
 pytest-asyncio>=0.10.0,<0.15.0
 pytest-cov>=2.7.0,<2.11.0
 openwisp-utils[qa]>=0.6.0,<0.8
+responses==0.12.1
```
nemesisdesign, line 7: It's better to use something like `responses~=0.12`

django_loci/tests/base/static/test-geocode.json
```diff
@@ -0,0 +1,23 @@
...
+            "attributes": {},
+            "extent": {
+                "xmin": 37.615020000000065,
+                "ymin": 55.749120000000055,
+                "xmax": 37.62502000000007,
+                "ymax": 55.75912000000006
+            }
+        }
+    ]
+}
```
nemesisdesign, line 23: please add a final blank line, applies to the other file as well

nemesisdesign (changes requested):
@Purhan can you isolate which test is failing? There may be another test in which we need to add mocking.

Purhan:
> @Purhan can you isolate which test is failing? There may be another test in which we need to add mocking.

@nemesisdesign Done, all tests pass without any exceptions now.

devkapilbansal (approved):
LGTM :+1: 
Tested locally and all test case passed while offline

nemesisdesign (approved):
Great work, thanks @Purhan @devkapilbansal :+1:

PR #86, opened 2021-03-12
[fix] Fixed not assigning of an existing geographic location #85

### LINKED ISSUE #85: [bug] Assigning an existing geographic location is broken
**How to replicate:**

- create a device
- create a geographic location in a separate step
- go to the device created in the first step, then go to map tab, select "Existing"
- select the geographic location just created
- hit save

**Expected result:**

- when the location is selected, the geo information should appear
- device is saved normally

**Actual result:**

- when the location is selected, the geo information does not appear
- device is not saved because a validation error

niteshsinha17:
Problem was that when we select an existing ``geo location``, ``locationChange`` was not triggered because input was not modified by user. So we need to listen some other event caused by user and which may change ``geo location``.

closes #85

### DIFF
--- a/django_loci/static/django-loci/css/loci.css
+++ b/django_loci/static/django-loci/css/loci.css
@@ -4,7 +4,6 @@
 /* hide redundant heading in objectlocationinline */
 .loci.coords > h2{ display: none }
 /* improve raw_id_field look */
-/* .coords .field-location input[type=text]{ display: none } */
 .field-location .related-lookup{
     width: auto;
     padding-left: 20px;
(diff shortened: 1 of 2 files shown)


django_loci/static/django-loci/css/loci.css
```diff
@@ -4,7 +4,6 @@
 /* hide redundant heading in objectlocationinline */
 .loci.coords > h2{ display: none }
 /* improve raw_id_field look */
-.coords .field-location input[type=text]{ display: none }
```
niteshsinha17, line 7: I am not sure why this css is used. Can you please confirm removing this is not affecting anything.

nemesisdesign (commented):
Thanks @niteshsinha17 you helped me understand the issue.
I found out why it stopped working: it's because a recent change in the django JS code.
I fixed it in https://github.com/openwisp/django-loci/commit/0774e0880df8d21fea15f076ff2ae28b1fb22de0.

PR #92, opened 2021-08-10
[deps] Update openwisp-utils

niteshsinha17:


.github/workflows/ci.yml
```diff
@@ -40,17 +40,17 @@ jobs:
         libsqlite3-mod-spatialite \
         gdal-bin
 
-    - name: Install python dependencies
-      run: pip install -r requirements-test.txt
-
```
nemesisdesign, line 45: I would avoid moving this and pin the version also here: https://github.com/openwisp/django-loci/blob/master/requirements-test.txt#L7

nemesisdesign (changes requested):
The build is failing.

devkapilbansal:
Fixed in #99

PR #97, opened 2021-10-13
[change] Convert geocoding test to check #90

### LINKED ISSUE #90: [change] Adjust or remove strict test
This [strict test](https://github.com/openwisp/django-loci#django-loci-geocode-strict-test) feature is backfiring quite a bit.

It's triggered every time django is initialized, also with scripts, management commands, celery, etc.

We should address this, maybe we can turn this into a management command to execute only some times.

devkapilbansal:
Closes #90

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -236,17 +236,17 @@ Amount of retry API calls when geocoding requests fail.
 
 API key if required (eg: Google Maps).
 
-Checks
-------
+System Checks
+-------------
 
 ``geocoding``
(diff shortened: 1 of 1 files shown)


pniaps:
Will this check be mandatory? I have a controller without internet

nemesisdesign:
> Will this check be mandatory? I have a controller without internet

Good point @pniaps, in the current version, this check is enabled by default but now we're making it optional.

PR #99, opened 2021-10-18
[deps] django version support

### LINKED ISSUE #94: [deps] Add support for django 3.2 and django 4.0 alpha
Let's make sure this module can run with Django 3.2 and Django 4.0 alpha.

queenish001:
Closes #94 
Added version support for django 3.2 and django 4.0 alpha in ci.yml


devkapilbansal (dismissed):
@queenish001 just one minor request here :point_down:

django_loci/base/admin.py
```diff
@@ -69,22 +69,22 @@ def get_urls(self):
...
             ),
-            re_path(
+            path(
                 r'^(?P<pk>[^/]+)/floorplans/json/$',
                 self.admin_site.admin_view(self.floorplans_json_view),
                 name='{0}_location_floorplans_json'.format(app_label),
             ),
-            re_path(
+            path(
                 r'^geocode/$',
```
ManishShah120, line 83: Hi, @queenish001 I guess we don't require, the url pattern to be in regex format for simple patterns when using `path`, can you look at it once. :smile:

django_loci/base/admin.py
```diff
@@ -69,22 +69,22 @@ def get_urls(self):
...
                 self.admin_site.admin_view(geocode_view),
                 name='{0}_location_geocode_api'.format(app_label),
             ),
-            re_path(
+            path(
                 r'^reverse-geocode/$',
```
ManishShah120, line 88: Same here and others.

ManishShah120 (changes requested):

django_loci/base/admin.py
nemesisdesign: if these works, do we really need to change all of them in all modules?
@pandafy @atb00ker @codesankalp what do you think?

codesankalp: We can use `re_path` instead of changing all the regex-based URL patterns in all modules.

queenish001: Um, So should I :
1. Change it back to `re_path`
Or
2. Keep it `path` and work on that regex issue?

nemesisdesign: let's do like it was done in openwisp-users and as Manish suggested initially, which looks cleaner and better for the long run: https://github.com/openwisp/openwisp-users/commit/289801e629cf969e5ea839dbf360c0914d443d01

Please do alwaysy some manual testing though, @queenish001, I ran the app and tried to open the admin page and I am getting 404, which hints to me you did the application after your changes which is a huge red flag and should never be done, always test the software you work on, not only with automated test suite but also with manual testing, and if how to do manual testing is not clear please ask.

django_loci/static/django-loci/js/loci.js
```diff
@@ -54,11 +54,11 @@ django.jQuery(function ($) {
     if (!gettext) { window.gettext = function (text) { return text; }; }
 
     function getLocationJsonUrl(pk) {
-        return baseLocationJsonUrl.replace('0000', pk);
+        return baseLocationJsonUrl.replace('00000000-0000-0000-0000-000000000000', pk);
```
devkapilbansal, line 57: Why this change is needed? :thinking:

.github/workflows/ci.yml
```diff
@@ -17,9 +17,14 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.7, 3.8, 3.9]
-        django-version: 
+        django-version:
           - django~=3.0.0
           - django~=3.1.0
+          - django~=3.2.0
+        include:
+          - django-version: django~=4.0a1
```
devkapilbansal, line 25:
```suggestion
          - django-version: django~=4.0b1
```

django_loci/static/django-loci/js/loci.js
pandafy: Yes, the previous url regex was not very strict for UUID. Since, I rewrote the url patterns using `path`, correct placeholder are required to match the url pattern with uuid.

.github/workflows/ci.yml
pandafy: Good catch, but we should now test with django~=4.0rc1

devkapilbansal:
Hi,
On investigating, I find out that `django-compress-staticfiles` doesn't plays well here. Uninstalling it is fixing issues on master branch of openwisp-utils

pandafy:
`channels~=2.0` is incompatible with Django 4. I added https://github.com/openwisp/django-loci/pull/99/commits/2c0bc3e2338e307b1756eae6b1f044a1ffaa6739 and https://github.com/openwisp/django-loci/pull/99/commits/67688a1ab3cb2459f8cf559cc529be2ee85a22dd in the PR for the sake of making CI build pass.

nemesisdesign (approved):

PR #108, opened 2022-09-08
[change] Use DEFAULT_STORAGE_CLASS as base for OverwriteStorage #107

### LINKED ISSUE #107: [change] Rethink LOCI_FLOORPLAN_STORAGE
Maybe it's overkill to have a dedicated storage class for django-loci, we could investigate whether we can default to the global storage class.

I believe the purpose of the custom storage class was to:

- delete the floorplan files when floorplan objects are deleted
- ensure uploading a new floorplan overwrites the current floorplan successfully

pandafy:
The DEFAULT_STORAGE_CLASS will be used for creating OverwriteStorage class.

Closes #107

### DIFF
--- a/django_loci/storage.py
+++ b/django_loci/storage.py
@@ -1,4 +1,7 @@
-from django.core.files.storage import FileSystemStorage
+from django.conf import settings
+from django.utils.module_loading import import_string
+
+DefaultFileStorageClass = import_string(getattr(settings, 'DEFAULT_FILE_STORAGE'))
 
 
 class OverwriteMixin:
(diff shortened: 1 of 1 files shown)


pandafy:
In my investigation I found out that using a custom backend for providing overwrite functionality is the best and simplest solution. I also came across [a related Django ticket](https://code.djangoproject.com/ticket/26563) which partly discusses this functionality. 

The problem with the current (master branch) implementation arises when we use a different storage backend (e.g. Cloud based storage backend) as the default file storage backend for Django. Even after configuring the `DEFAULT_FILE_STORAGE` setting, Django Loci keeps using the `FileSystemStorage`. In this PR, I have updated the code to add the overwriting functionality to the file storage class defined in the `DEFAULT_FILE_STORAGE` setting.

nemesisdesign (approved):

PR #157, opened 2025-03-13
[admin] Fixed 'invalid value' error on location type change #156

### LINKED ISSUE #156: [bug] Vague "invalid value" validation error
**Describe the bug**
Vague 'invalid error' was being thrown due to type of location not being synchronized with type change in form. Also user was allowed to change type of location from indoor to outdoor without deleting related floorplans.

**Steps To Reproduce**
1. Login to admin dashboard
2. Create an outdoor location
3. Create a device, assign the outdoor location created in the previous step, save
4. Now open the device again, change the outdoor location to be indoor, add a new floorplan, specify floor number, specify an image, set a point on the image (to specify indoor coordinates), then save
5. Alert 'Please correct the error below' is displayed with no reference as to what happened.

Another way to reproduce the behavior (on openwisp-controller):
1. Create an indoor location, lets say 'Indoor'
2. Create a device, assign the indoor location created in the previous step, save
3. Create an outdoor location, lets say 'Outdoor'
4. Now open the device again, change the location from 'Indoor' to 'Outdoor', then save
5. Alert 'Please correct the error below' is displayed with no reference as to what happened.

**Expected behavior**
It should get saved properly and also proper alert to be shown when switching indoor to outdoor within same location

**Screenshots**

Image

**System Informatioon:**
 - OS: MacOS Sequoia 15.3.1
 - Python Version: 3.12.7
 - Django Version: 4.2.20
 - Browser Version: 1.73.105

DragnEmperor:
## Checklist

- [x] I have written new test cases for new code and/or updated existing tests for changes to existing code.

Closes #156 

## Description of Changes
Synchronize location type and clean 'indoor' value in clean function of 'AbstractObjectLocationForm'. 
Also added restriction to user from changing type from indoor to outdoor by adding 'floorplans' checks.

### DIFF
--- a/django_loci/base/admin.py
+++ b/django_loci/base/admin.py
@@ -289,6 +289,10 @@ class AbstractObjectLocationForm(forms.ModelForm):
         fields = []
         if not is_mobile and type_ in ['outdoor', 'indoor']:
             fields += ['location_selection', 'name', 'address', 'geometry']
+        # sync location and clean indoor field basis type
+        if location := data.get('location'):
+            location.type = type_
+            data['indoor'] = None if type_ != 'indoor' else data.get('indoor')
         if type_ == 'indoor':
             if data.get('floorplan_selection') == 'existing':
                 fields.append('floorplan')
--- a/django_loci/base/admin.py
+++ b/django_loci/base/admin.py
@@ -72,9 +72,15 @@ class AbstractLocationForm(forms.ModelForm):
         css = {'all': ('django-loci/css/loci.css',)}
 
     def clean(self):
-        # if changing existing location to outdoor, delete all floorplans
-        if self.instance and self.cleaned_data.get('type') == "outdoor":
-            self.instance.type = self.cleaned_data['type']
+        location_type = self.cleaned_data.get('type')
+        instance_type = self.instance.type
+        # if 'type' is not None implies that location is being changed, hence delete floorplans
+        if (
+            location_type == 'outdoor'
+            and instance_type
+            and instance_type != location_type
+        ):
+            self.instance.objectlocation_set.all().update(floorplan=None)
             self.instance.floorplan_set.all().delete()
 
 
@@ -298,7 +304,7 @@ class AbstractObjectLocationForm(forms.ModelForm):
         # sync location, clean indoor field basis type and delete floorplans if type is not outdoor
         if location := data.get('location'):
             location.type = type_
-            if type_ != "indoor":
+            if type_ != 'indoor':
                 data['indoor'] = None
                 # set floorplans of all objectlocations to None where this location is related
                 location.objectlocation_set.update(floorplan=None)
--- a/django_loci/base/admin.py
+++ b/django_loci/base/admin.py
@@ -71,18 +71,6 @@ class AbstractLocationForm(forms.ModelForm):
         )
         css = {'all': ('django-loci/css/loci.css',)}
 
-    def clean(self):
-        location_type = self.cleaned_data.get('type')
-        instance_type = self.instance.type
-        # if 'type' is not None implies that location is being changed, hence delete floorplans
-        if (
-            location_type == 'outdoor'
-            and instance_type
-            and instance_type != location_type
-        ):
-            self.instance.objectlocation_set.all().update(floorplan=None)
-            self.instance.floorplan_set.all().delete()
-
 
 class AbstractFloorPlanInline(TimeReadonlyAdminMixin, admin.StackedInline):
     extra = 0
@@ -301,14 +289,10 @@ class AbstractObjectLocationForm(forms.ModelForm):
         fields = []
         if not is_mobile and type_ in ['outdoor', 'indoor']:
             fields += ['location_selection', 'name', 'address', 'geometry']
-        # sync location, clean indoor field basis type and delete floorplans if type is not outdoor
+        # sync location, clean indoor field basis type
         if location := data.get('location'):
             location.type = type_
-            if type_ != 'indoor':
-                data['indoor'] = None
-                # set floorplans of all objectlocations to None where this location is related
-                location.objectlocation_set.update(floorplan=None)
-                location.floorplan_set.all().delete()
+            data['indoor'] = None if type_ != 'indoor' else data.get('indoor')
         if type_ == 'indoor':
             if data.get('floorplan_selection') == 'existing':
                 fields.append('floorplan')
--- a/django_loci/base/admin.py
+++ b/django_loci/base/admin.py
@@ -71,6 +71,12 @@ class AbstractLocationForm(forms.ModelForm):
         )
         css = {'all': ('django-loci/css/loci.css',)}
 
+    def clean(self):
+        # if changing existing location to outdoor, delete all floorplans
+        if self.instance and self.cleaned_data.get('type') == "outdoor":
+            self.instance.type = self.cleaned_data['type']
+            self.instance.floorplan_set.all().delete()
+
 
 class AbstractFloorPlanInline(TimeReadonlyAdminMixin, admin.StackedInline):
     extra = 0
@@ -289,10 +295,14 @@ class AbstractObjectLocationForm(forms.ModelForm):
         fields = []
         if not is_mobile and type_ in ['outdoor', 'indoor']:
             fields += ['location_selection', 'name', 'address', 'geometry']
-        # sync location and clean indoor field basis type
+        # sync location, clean indoor field basis type and delete floorplans if type is not outdoor
         if location := data.get('location'):
             location.type = type_
-            data['indoor'] = None if type_ != 'indoor' else data.get('indoor')
+            if type_ != "indoor":
+                data['indoor'] = None
+                # set floorplans of all objectlocations to None where this location is related
+                location.objectlocation_set.update(floorplan=None)
+                location.floorplan_set.all().delete()
         if type_ == 'indoor':
             if data.get('floorplan_selection') == 'existing':
                 fields.append('floorplan')
--- a/django_loci/base/models.py
+++ b/django_loci/base/models.py
@@ -75,6 +75,7 @@ class AbstractLocation(TimeStampedEditableModel):
         # if location type is changed to outdoor, remove all associated floorplans
         if (
             self.type != self._initial_type
+            and not self._state.adding
             and self.type == 'outdoor'
             and self.floorplan_set.exists()
         ):
--- a/django_loci/base/models.py
+++ b/django_loci/base/models.py
@@ -5,7 +5,6 @@ from django.contrib.contenttypes.models import ContentType
 from django.contrib.gis.db import models
 from django.contrib.humanize.templatetags.humanize import ordinal
 from django.core.exceptions import ValidationError
-from django.db import transaction
 from django.utils.translation import gettext_lazy as _
 
 from openwisp_utils.base import TimeStampedEditableModel
@@ -48,6 +47,11 @@ class AbstractLocation(TimeStampedEditableModel):
     class Meta:
         abstract = True
 
+    # overriding __init__ to store the initial type
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._initial_type = self.type
+
     def __str__(self):
         return self.name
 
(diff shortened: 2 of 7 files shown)


niteshsinha17:
@DragnEmperor Thanks. I tested this locally, There seems a bug.

Steps:
1. Create a location
2. Select type indoor
3. Add floor and fill address
4. change type to outdoor
5. submit and see there is a validation error but none of the fields are highlighted.

There are multiple ways to get the same type of error from an inline, create, or edit form. I see a JS file has been added to handle this case in the edit form. 

I suggest sticking to the same approach either from js code or Python code.

django_loci/static/django-loci/js/loci.js
```diff
@@ -198,6 +199,15 @@ django.jQuery(function ($) {
       $geoEdit.hide();
       $indoor.hide();
       $typeRow.show();
+    }
+      // Show alert when switching from indoor to outdoor if floorplans exist
+    if (value === "outdoor" && floorplans_length > 1){
+      var msg =
+        "Please remove the associated floorplans first " +
+        'and save; then you can switch to type "indoor"';
+      alert(gettext(msg));
```
nemesifier, line 208: I have an idea which may be better. What if instead of using an alert and ask the user to do this action, we do it for them after asking for confirmation with a confirm dialog? If they hit yes we remove the floorplan and hit the save with JS. Better UX. What do you think @niteshsinha17?

niteshsinha17: @nemesifier Please confirm, This also means if the user has changed the device name and trying to change the type and hits yes this will lead to saving changes of device name in the db. If yes, then I think it's not a better UX.

I think this can be a better option,

Disable editing of type and show some help text to remove floor plans first to change the type. This can be done via both JS and Python code.

What do you think?

niteshsinha17: Also, in inline forms, js is not running, 

This js code has been added to show an alert when a user tries to change the type but I can't find its log in the console.

https://github.com/openwisp/django-loci/blob/master/django_loci/static/django-loci/js/floorplan-inlines.js

DragnEmperor: @niteshsinha17 yes in the current code floorplan-inlines.js is used only for location form.

DragnEmperor: Should we do this for normal location form too?

DragnEmperor: @nemesifier  I have added a commit to reflect this. For normal location form, i dont have access to floorplan instance in location form so i handled it via LocationAdmin to disable its validation/creation when location is outdoor. I have kept this handling mainly on python code and removed js code for location change alert and any other previous code which was added by me in js. Also updated the test cases for the same

django_loci/base/admin.py
```diff
@@ -289,6 +295,14 @@ def clean(self):
...
         if not is_mobile and type_ in ['outdoor', 'indoor']:
             fields += ['location_selection', 'name', 'address', 'geometry']
+        # sync location, clean indoor field basis type and delete floorplans if type is not outdoor
+        if location := data.get('location'):
+            location.type = type_
+            if type_ != "indoor":
+                data['indoor'] = None
+                # set floorplans of all objectlocations to None where this location is related
+                location.objectlocation_set.update(floorplan=None)
+                location.floorplan_set.all().delete()
```
niteshsinha17, line 305: This will delete all floors associated with an indoor location.

Consider this scenario:
An indoor location has two floors, and two different devices are linked to these locations. When the type of one device is changed, it also affects the type of the other device. I have tested and confirmed this behavior.

@nemesifier, what do you think? Is this expected behavior?

Upon further investigation, I found that the main issue occurs when a floor is linked to an outdoor location (floors should only be linked to indoor locations). This causes an error, which happens specifically when changing the location type through the device. Attaching a video for reference. 

https://github.com/user-attachments/assets/f5b83cde-9944-448f-8efc-632b0bcf68e4

niteshsinha17: I still think, type of location should not be editable, If user want to change type they can select a new location or remove Geographic information.

DragnEmperor: yes @niteshsinha17 i also encountered this issue during initial commits , when a floor references an outdoor location when changed via devicelocation form so I added a restriction basis the one in location form to alert the user of associated floorplans. 
As you said we can approach this two ways:
- deleting all related floorplans (affects other devices associated to the same location), or
- make the type field non editable/ restriction on change for associated floorplans in device location form.

django_loci/base/admin.py
```diff
@@ -71,6 +71,12 @@ class Media:
         )
         css = {'all': ('django-loci/css/loci.css',)}
 
+    def clean(self):
+        # if changing existing location to outdoor, delete all floorplans
+        if self.instance and self.cleaned_data.get('type') == "outdoor":
+            self.instance.type = self.cleaned_data['type']
+            self.instance.floorplan_set.all().delete()
+
```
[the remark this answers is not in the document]
DragnEmperor: Yes. My bad, working on it

django_loci/base/admin.py
DragnEmperor: Also had a doubt, why arent we deleting objectlocation models on deletion of devices?

django_loci/base/admin.py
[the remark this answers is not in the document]
DragnEmperor: @niteshsinha17 i tried keeping it in location model's save, but it is not reached when changing type in Location form as clean method of location model raises error: 'this location has floorplans associated to it, please delete them before changing its type'
Screenshot 2025-03-21 at 11 08 46 AM
I think the flow of execution is 'form-clean -> model-clean -> form-save -> model-save
I missed seting related objectlocations floorplans to null before deleting related floorplans in location form. Will add it.

Also the above validation error is not raised when changing object via device admin as the instances for which it runs 'clean' method are itself and the inline instance i.e. device model and object location model.

django_loci/base/admin.py
niteshsinha17: As we have multiple ways to solve this issue @nemesifier can suggest which one to implement in better ways. Lets wait for his response.

django_loci/base/admin.py
```diff
@@ -71,6 +71,12 @@ class Media:
         )
         css = {'all': ('django-loci/css/loci.css',)}
 
+    def clean(self):
+        # if changing existing location to outdoor, delete all floorplans
+        if self.instance and self.cleaned_data.get('type') == "outdoor":
```
nemesifier, line 76:
```suggestion
        if self.instance and self.cleaned_data.get('type') == 'outdoor':
```
Please be consistent with the quoting style of strings.

django_loci/base/admin.py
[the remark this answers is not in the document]
nemesifier: Is there a way to check whether the type is being changed from indoor to outdoor before proceeding with such a change?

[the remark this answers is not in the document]
DragnEmperor: Yes. Added in the new commit

nemesifier (changes requested):
I thought I had left some review comments yesterday but I must have done something wrong (sorry for that!).

I think we need to let the user know and confirm (eg: a JS confirm dialog) that the type is being changed from indoor to outdoor and this will delete any floorplan data associated to the location, which may affect other devices if they share the same location.

Moreover, I think we need to implement this delete operation with a model method because I think we need to clean up any floorplan position data, eg:

- Not only we need to delete the floorplans of a location which switches from indoor to outdoor.
- Any object relating to the floorplan being deleted or location changing from indoor to outdoor should have indoor coordinates set, we need to clear these as well to avoid data inconsistency, probably this needs to be done before deleting the floorplans.
- We should do both operations in a [DB transaction](https://docs.djangoproject.com/en/5.1/topics/db/transactions/) to avoid ending up in an inconsistent state in case anything fails.

niteshsinha17:
> I thought I had left some review comments yesterday but I must have done something wrong (sorry for that!).
> 
> I think we need to let the user know and confirm (eg: a JS confirm dialog) that the type is being changed from indoor to outdoor and this will delete any floorplan data associated to the location, which may affect other devices if they share the same location.
> 
> Moreover, I think we need to implement this delete operation with a model method because I think we need to clean up any floorplan position data, eg:
> 
> * Not only we need to delete the floorplans of a location which switches from indoor to outdoor.
> * Any object relating to the floorplan being deleted or location changing from indoor to outdoor should have indoor coordinates set, we need to clear these as well to avoid data inconsistency, probably this needs to be done before deleting the floorplans.
> * We should do both operations in a [DB transaction](https://docs.djangoproject.com/en/5.1/topics/db/transactions/) to avoid ending up in an inconsistent state in case anything fails.

I was also concern about these points. Thanks

DragnEmperor:
> I thought I had left some review comments yesterday but I must have done something wrong (sorry for that!).
> 
> I think we need to let the user know and confirm (eg: a JS confirm dialog) that the type is being changed from indoor to outdoor and this will delete any floorplan data associated to the location, which may affect other devices if they share the same location.
> 
> Moreover, I think we need to implement this delete operation with a model method because I think we need to clean up any floorplan position data, eg:
> 
> * Not only we need to delete the floorplans of a location which switches from indoor to outdoor.
> * Any object relating to the floorplan being deleted or location changing from indoor to outdoor should have indoor coordinates set, we need to clear these as well to avoid data inconsistency, probably this needs to be done before deleting the floorplans.
> * We should do both operations in a [DB transaction](https://docs.djangoproject.com/en/5.1/topics/db/transactions/) to avoid ending up in an inconsistent state in case anything fails.

@nemesifier Thanks for the review. Added new commit reflecting the required changes

django_loci/base/models.py
```diff
@@ -81,6 +66,14 @@ def _validate_geometry_if_not_mobile(self):
     def short_type(self):
         return _(self.type.capitalize())
 
+    def save(self, *args, **kwargs):
+        # if location type is changed to outdoor, remove all associated floorplans
+        if self.type == 'outdoor' and not self._state.adding:
+            with transaction.atomic():
+                self.objectlocation_set.all().update(floorplan=None, indoor=None)
+                self.floorplan_set.all().delete()
+        return super().save(*args, **kwargs)
```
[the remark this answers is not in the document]
DragnEmperor: Since save method is wrapped in transaction itself so i am removing this.

django_loci/base/models.py
```diff
@@ -66,12 +70,16 @@ def _validate_geometry_if_not_mobile(self):
...
 
+    # save method is automatically wrapped in atomic transaction
     def save(self, *args, **kwargs):
         # if location type is changed to outdoor, remove all associated floorplans
-        if self.type == 'outdoor' and not self._state.adding:
-            with transaction.atomic():
-                self.objectlocation_set.all().update(floorplan=None, indoor=None)
-                self.floorplan_set.all().delete()
+        if (
+            self.type != self._initial_type
```
nemesifier, line 77: Let's add also a condition which ensures this is not executed on new objects, should be `not self._state.adding` or something similar.

nemesifier (changes requested):

django_loci/base/models.py
DragnEmperor: I thought self.type != self._initial_type should do trick but for new objects also this will be true as _initial_type will be None. Adding the `self._state.adding` check back. Thanks

nemesifier (approved):

django_loci/tests/base/test_admin_inline.py
```diff
@@ -776,3 +776,121 @@ def test_add_indoor_location_without_floor(self):
...
+        self._login_as_admin()
+        p = 'floorplan_set'
+        floorplan_file = open(self._floorplan_path, 'rb')
+        params = {
+            'name': 'test-add-outdoor-with-floorplan',
+            'type': 'outdoor',
+            'geometry': 'SRID=4326;POINT (12.512324 41.898703)',
+            'address': 'Piazza Venezia, Roma, Italia',
+            '{0}-0-floor'.format(p): '1',
+            '{0}-0-image'.format(p): floorplan_file,
```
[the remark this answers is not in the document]
DragnEmperor: Thanks @pandafy . Have made the required changes.

nemesifier (approved):
Great work @DragnEmperor, thanks! :+1:

PR #170, opened 2025-04-11
[chores] Migrate formset function to base objectlocation mixin

DragnEmperor:
## Checklist

## Description of Changes

The get_formset_kwargs method was previously part of AbstractObjectLocationInline, which is specific to models using generic foreign keys.
Moving it to ObjectLocationMixin ensures the method is available to all models, improving reusability and consistency.

### DIFF
--- a/django_loci/tests/__init__.py
+++ b/django_loci/tests/__init__.py
@@ -96,9 +96,6 @@ class TestAdminMixin(object):
 
     def _create_readonly_admin(self, **kwargs):
         """Creates a read-only admin user with view permissions for the specified models."""
-        # during test discovery 'auth' app is not ready so import here
-        from django.contrib.auth.models import Permission
-
         models = kwargs.pop('models', [])
         user = self._create_admin(is_superuser=False, **kwargs)
         if models:
@@ -106,7 +103,7 @@ class TestAdminMixin(object):
             for model in models:
                 permission_codenames.append(f'view_{model.__name__.lower()}')
             # assign view permissions to user
-            view_permission = Permission.objects.filter(
+            view_permission = self.permission_model.objects.filter(
                 codename__in=permission_codenames
             )
             user.user_permissions.add(*view_permission)
--- a/django_loci/tests/__init__.py
+++ b/django_loci/tests/__init__.py
@@ -96,16 +96,20 @@ class TestAdminMixin(object):
 
     def _create_readonly_admin(self, **kwargs):
         """Creates a read-only admin user with view permissions for the specified models."""
-        #
+        # during test discovery 'auth' app is not ready so import here
         from django.contrib.auth.models import Permission
 
-        permission_codename = []
-        for model in kwargs.pop('models', []):
-            permission_codename.append(f'view_{model.__name__.lower()}')
-        # assign view permissions to user
-        view_permission = Permission.objects.filter(codename__in=permission_codename)
+        models = kwargs.pop('models', [])
         user = self._create_admin(is_superuser=False, **kwargs)
-        user.user_permissions.add(*view_permission)
+        if models:
+            permission_codenames = []
+            for model in models:
+                permission_codenames.append(f'view_{model.__name__.lower()}')
+            # assign view permissions to user
+            view_permission = Permission.objects.filter(
+                codename__in=permission_codenames
+            )
+            user.user_permissions.add(*view_permission)
         return user
 
     def _load_content(self, file):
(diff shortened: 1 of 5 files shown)


codesankalp (changes requested):
@DragnEmperor I approved the tests workflow, but they’re still failing. If the failures are due to a dependency on your django-loci branch, please add a tarball link to your branch in requirements.txt to ensure the tests pass.

Once the django-loci PR is merged, you can revert the requirements.txt change in this PR.

django_loci/tests/__init__.py
```diff
@@ -94,6 +94,20 @@ def _login_as_admin(self):
         self.client.force_login(admin)
         return admin
 
+    def _create_readonly_admin(self, **kwargs):
+        """Creates a read-only admin user with view permissions for the specified models."""
+        #
+        from django.contrib.auth.models import Permission
```
codesankalp, line 100: Do we need this import here or can it be moved to top?

django_loci/tests/__init__.py
```diff
@@ -94,6 +94,20 @@ def _login_as_admin(self):
...
+
+        permission_codename = []
+        for model in kwargs.pop('models', []):
+            permission_codename.append(f'view_{model.__name__.lower()}')
+        # assign view permissions to user
+        view_permission = Permission.objects.filter(codename__in=permission_codename)
+        user = self._create_admin(is_superuser=False, **kwargs)
+        user.user_permissions.add(*view_permission)
```
codesankalp, line 108: The user should be created first, and then permissions can be assigned. This is because the kwargs models list can be empty, and adding view permissions is not always necessary when the queryset is empty.

codesankalp: One more question - do we really need to pass models to this function? Can’t we automatically assign all read-only model permissions to the user instead?

django_loci/tests/base/test_selenium.py
```diff
@@ -11,10 +11,13 @@
...
     SeleniumTestMixin, TestAdminInlineMixin, TestLociMixin
 ):
+    def _create_device(self):
+        self.find_element(by=By.NAME, value='name').send_keys('11:22:33:44:55:66')
+
     def test_create_new_device(self):
         self.login()
         self.open(reverse(self.add_url))
-        self.find_element(by=By.NAME, value='name').send_keys('11:22:33:44:55:66')
+        self._create_device()
```
codesankalp, line 20: Is there a specific reason for creating this new one-line function? I also noticed that it’s not being used in multiple places. Also as it is just filling the form shoul we need to name this function `_create_device`?

codesankalp (changes requested):

django_loci/tests/__init__.py
DragnEmperor: I was getting `Apps aren't loaded yet.`. I think it is because during test discovery, apps are not loaded.

django_loci/tests/base/test_selenium.py
DragnEmperor: for device admin in `openwisp-controller` there is another required field - mac address. So i just kind of kept the logic for device specific fields in another function to avoid rewriting the complete for `openwisp-controller`.

django_loci/tests/__init__.py
DragnEmperor: > One more question - do we really need to pass models to this function? Can’t we automatically assign all read-only model permissions to the user instead?

I was thinking of providing a fine grained control over permissions so as to simulate the working for a user with view-only access to specific models. Let me know if this requires any changes

django_loci/tests/__init__.py
devkapilbansal: Would be good to add a comment for that for future references

DragnEmperor: Comment Added

django_loci/tests/base/test_selenium.py
DragnEmperor: Renamed the function to `_fill_device_form`

nemesifier: @DragnEmperor there's no point in creating methods if the logic is not shared across different parts of the code or unless the logic is too big and needs to be split to increase readability.

nemesifier (approved):
Pushed some minor adjustments in https://github.com/openwisp/django-loci/pull/170/commits/60808270c1b635e077c6c9a6376c943dc63347eb and https://github.com/openwisp/django-loci/pull/170/commits/439ae91b4adb52fc996fa5edc54d35bb769cd160.

PR #171, opened 2025-04-27
[fix] Added address field to real-time location updates #169

### LINKED ISSUE #169: [bug] Real-time update logic does not update address field
I am not 100% sure if this is intentional or not, but I just noticed that the real-time update logic does not update address field:

https://github.com/user-attachments/assets/e542c641-28a8-4b55-bef5-13e2e3c51426

I believe that if the address field changes it would be a lot better to send this information in the update broadcasted via websocket, the UI then can decide what to do with it.

In the case of these two pages, I think we should update the address field accordingly.

We need a selenium browser test for this.
The [channels docs has some references on selenium testing](https://channels.readthedocs.io/en/stable/tutorial/part_4.html).
In the test, let's verify that both the coordinates and address are updated.

**System Informatioon:**
 - Python Version: 3.10
 - Django Version: Django 4.2.20

dee077:
Added address update handling in the websocket message and updated listenForLocationUpdates logic accordingly.

Fixes #169

## Checklist

- [ ] I have written new test cases for new code and/or updated existing tests for changes to existing code.

Closes #169.

## Description of Changes

- Updated the real-time update logic to include address field updates via WebSocket.
- Modified the `listenForLocationUpdates` function to handle and broadcast address changes.

## Issues

- Inheriting from `ChannelLiveServerTestCase` is causing test failures in Django 5.2, which is a new release, and a fix has yet to be released for it.

- Adding `ChannelLiveServerTestCase` prevents running tests in parallel with the --parallel flag, as it relies on live WebSocket connections that cannot run concurrently.

## Demo

[real-time-address.webm](https://github.com/user-attachments/assets/9cec1fda-b51e-4db9-8e32-cc02050fee20)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
       with:
         parallel: true
         format: cobertura
-        flag-name: python-${{ matrix.env.env }}
+        flag-name: python-${{ matrix.python-version }}-${{ matrix.django-version }}
         github-token: ${{ secrets.GITHUB_TOKEN }}
 
   coveralls:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,11 +86,7 @@ jobs:
 
     - name: Tests
       if: ${{ !cancelled() && steps.deps.conclusion == 'success' }}
-      run: |
-        coverage run runtests.py --parallel --noinput --exclude-tag=selenium_tests
-        coverage run runtests.py --noinput --tag=selenium_tests
-        coverage combine
-        coverage xml
+      run: ./runtests
       env:
         SELENIUM_HEADLESS: 1
         GECKO_LOG: 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,8 @@ jobs:
     - name: Tests
       if: ${{ !cancelled() && steps.deps.conclusion == 'success' }}
       run: |
-        coverage run runtests.py --parallel
+        coverage run runtests.py --parallel --exclude-tag=selenium_tests
+        coverage run runtests.py --tag=selenium_tests
         coverage combine
         coverage xml
       env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,8 +87,8 @@ jobs:
     - name: Tests
       if: ${{ !cancelled() && steps.deps.conclusion == 'success' }}
       run: |
-        coverage run runtests.py --parallel --exclude-tag=selenium_tests
-        coverage run runtests.py --tag=selenium_tests
+        coverage run runtests.py --parallel --noinput --exclude-tag=selenium_tests
+        coverage run runtests.py --noinput --tag=selenium_tests
         coverage combine
         coverage xml
       env:
--- a/django_loci/tests/base/test_channels.py
+++ b/django_loci/tests/base/test_channels.py
@@ -143,7 +143,11 @@ class BaseTestChannels(TestAdminMixin, TestLociMixin):
         assert connected
         await self._save_location(request_vars["pk"])
         response = await communicator.receive_json_from()
-        assert response == {"type": "Point", "coordinates": [12.513124, 41.897903]}
+        print(response)
+        assert response == {
+            "geometry": {"type": "Point", "coordinates": [12.513124, 41.897903]},
+            "address": "Via del Corso, Roma, Italia",
+        }
         await communicator.disconnect()
 
     async def _save_location(self, pk):
--- a/django_loci/tests/base/test_channels.py
+++ b/django_loci/tests/base/test_channels.py
@@ -143,7 +143,6 @@ class BaseTestChannels(TestAdminMixin, TestLociMixin):
         assert connected
         await self._save_location(request_vars["pk"])
         response = await communicator.receive_json_from()
-        print(response)
         assert response == {
             "geometry": {"type": "Point", "coordinates": [12.513124, 41.897903]},
             "address": "Via del Corso, Roma, Italia",
--- a/tests/openwisp2/settings.py
+++ b/tests/openwisp2/settings.py
@@ -12,6 +12,9 @@ DATABASES = {
     "default": {
         "ENGINE": "openwisp_utils.db.backends.spatialite",
         "NAME": "django-loci.db",
+        "TEST": {
+            "NAME": "openwisp_utils_test.db",
+        },
     }
 }
 
--- a/tests/openwisp2/settings.py
+++ b/tests/openwisp2/settings.py
@@ -16,7 +16,7 @@ DATABASES = {
 }
 if TESTING and "--exclude-tag=selenium_tests" not in sys.argv:
     DATABASES["default"]["TEST"] = {
-        "NAME": os.path.join(BASE_DIR, "django-loci.db.db"),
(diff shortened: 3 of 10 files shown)


django_loci/tests/base/test_channels.py
```diff
@@ -143,7 +143,11 @@ async def test_location_update(self):
         assert connected
         await self._save_location(request_vars['pk'])
         response = await communicator.receive_json_from()
-        assert response == {'type': 'Point', 'coordinates': [12.513124, 41.897903]}
+        print(response)
```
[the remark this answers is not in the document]
dee077: My bad! It got missed

pandafy (commented):
I installed this PR with openwisp-controller and performed the following test: 

1. Opened a location admin in the browser and ensured that the websocket coonnection is working
2. Updated the `geometry` field of the location using [the REST API](https://openwisp.io/docs/dev/controller/user/rest-api.html#change-location-details) with a PATCH request. 

**Expected outcome**
The location admin updated the pin on the map according to the received data from websocket. But, the `address` field was still showing the old value. 

This occurs because the JS updates the address field in Django admin if the user changes the pin on the map.

https://github.com/openwisp/django-loci/blob/c6707ef99bf45a9f840030f1b8a096c34614aded/django_loci/static/django-loci/js/loci.js#L508-L529

But, the backend does not perform such operation on save. @nemesifier is this intentional? If so, shall we update the logic in this PR to update the address field using the JS?

nemesifier:
> I installed this PR with openwisp-controller and performed the following test:
> 
> 1. Opened a location admin in the browser and ensured that the websocket coonnection is working
> 2. Updated the `geometry` field of the location using [the REST API](https://openwisp.io/docs/dev/controller/user/rest-api.html#change-location-details) with a PATCH request.
> 
> **Expected outcome** The location admin updated the pin on the map according to the received data from websocket. But, the `address` field was still showing the old value.
> 
> This occurs because the JS updates the address field in Django admin if the user changes the pin on the map.
> 
> https://github.com/openwisp/django-loci/blob/c6707ef99bf45a9f840030f1b8a096c34614aded/django_loci/static/django-loci/js/loci.js#L508-L529
> 
> But, the backend does not perform such operation on save. @nemesifier is this intentional? If so, shall we update the logic in this PR to update the address field using the JS?

@pandafy I think the REST API is a different matter and we could discuss that separately. I am not sure now whether we would want to do that automatically or not

pandafy:
> @pandafy I think the REST API is a different matter and we could discuss that separately. I am not sure now whether we would want to do that automatically or not

Let's say a mobile device updated it's coordinates. The pin on the map will get updated, but the address field will keep showing the old value. Maybe, we can trigger an update to the address field in the admin using the JS logic.

nemesifier:
> > @pandafy I think the REST API is a different matter and we could discuss that separately. I am not sure now whether we would want to do that automatically or not
> 
> Let's say a mobile device updated it's coordinates. The pin on the map will get updated, but the address field will keep showing the old value. Maybe, we can trigger an update to the address field in the admin using the JS logic.

It's a good point.

A device which moves around should be able to ensure the address field is either updated or cleared out, depending on their implementation.

In this change, we are focusing on the admin to make the user experience consistent when using the web admin, the REST API has a different purpose and I believe it's the responsibility of the API client to decide what to do with the address field (clear it, update it or leave it as is), so I wouldn't involve the REST API in this change, unless I am missing something, if that's the case let me know.

In the video I recorded when creating the issue I flagged the "is mobile" checkbox without thinking about it, but I am more concerned with regular static locations, which usually have the address field filled, while mobile locations usually don't have it because they keep moving around.

I think we could tweak the code to update the address field only if it's not empty, that should be enough to not mess with mobile locations.

What do you think @dee077 @pandafy?

dee077:
## Updates:
- Added selenium tests.
- Updated ci to run unit tests and selenium tests separately.

.github/workflows/ci.yml
```diff
@@ -87,7 +87,8 @@ jobs:
     - name: Tests
       if: ${{ !cancelled() && steps.deps.conclusion == 'success' }}
       run: |
-        coverage run runtests.py --parallel
+        coverage run runtests.py --parallel --noinput --exclude-tag=selenium_tests
+        coverage run runtests.py --noinput --tag=selenium_tests
```
nemesifier, line 91: See https://github.com/openwisp/openwisp-utils/pull/506/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL88-R88

Let's make sure to update the README section which explains how to run tests.

tests/openwisp2/settings.py
```diff
@@ -12,6 +12,9 @@
     "default": {
         "ENGINE": "openwisp_utils.db.backends.spatialite",
         "NAME": "django-loci.db",
+        "TEST": {
+            "NAME": "openwisp_utils_test.db",
+        },
```
nemesifier, line 17: https://github.com/openwisp/openwisp-utils/pull/506/files#diff-26cea5c755fb048b49985aa06de44b90bfdfe6e9165f73f08bac2197df7e61e8R16-R19

nemesifier (changes requested):
@dee077 please see the changes I made to other modules regarding tests:

- https://github.com/openwisp/openwisp-utils/pull/506
- https://github.com/openwisp/openwisp-network-topology/pull/265
- https://github.com/openwisp/openwisp-firmware-upgrader/pull/340

Let's adopt a similar strategy, I will leave more pointers below.

.github/workflows/ci.yml
dee077: Correct commands are present in the docs!

tests/openwisp2/settings.py
dee077: Updated!

dee077:
## Updates
- Fixed test db logic as suggested
- Fixed the way of running tests in ci by adding a script `runtests`, like other modules.

nemesifier (dismissed):

nemesifier (commented):
@dee077 I found some problems during testing.

The selenium test logic was failing on my system 4 times, but passing at the 5.

```
--------------------------------------------------------------------------------
[Retry] Retrying "django_loci.tests.test_selenium.TestDeviceAdminSelenium.test_real_time_update_address_field", attempt 1/5. 
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
[Retry] Retrying "django_loci.tests.test_selenium.TestDeviceAdminSelenium.test_real_time_update_address_field", attempt 2/5. 
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
[Retry] Retrying "django_loci.tests.test_selenium.TestDeviceAdminSelenium.test_real_time_update_address_field", attempt 3/5. 
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
[Retry] Retrying "django_loci.tests.test_selenium.TestDeviceAdminSelenium.test_real_time_update_address_field", attempt 4/5. 
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
[Retry] Retrying "django_loci.tests.test_selenium.TestDeviceAdminSelenium.test_real_time_update_address_field", attempt 5/5. 
--------------------------------------------------------------------------------
```

It had to do with tabs, I've seen this before. For some reason the browser may start with an emtpy tab. I tried a fix 716acb8, let's see if the build passes.

Another thing: when calling `./runtests`, pytests is run both times that `runtests.py` is called, which is redundant. Plus you forgot to update the instructions on running tests in the README. I fixed that in a1ba01d.

PR #210, opened 2026-03-26
fix(security): enforce object-level permissions on admin JSON endpoints

phusi319:
## Summary
This PR adds explicit object-level authorization checks to the custom admin JSON endpoints for locations.

## Security issue
The custom admin endpoints:
- `/admin/.../location/<uuid>/json/`
- `/admin/.../location/<uuid>/floorplans/json/`

retrieved objects and returned data without checking `has_view_or_change_permission` for the target instance.

Even though these views are wrapped with `admin_view`, they should still enforce object-level checks before returning potentially sensitive location/floorplan data.

## Fix
- In `django_loci/base/admin.py`:
  - import `PermissionDenied`
  - add `has_view_or_change_permission(request, obj=instance)` checks in:
    - `json_view`
    - `floorplans_json_view`
  - raise `PermissionDenied` when unauthorized

- In `django_loci/tests/base/test_admin.py`:
  - add regression tests verifying users without location view permission receive `403` on both JSON endpoints.

## Notes
Could not fully execute project tests locally because this environment misses OpenWISP runtime dependencies (`openwisp2`) required by the test setup. Added focused regression tests for CI to validate behavior.

Signed-off-by: Phu Si On <phusi319@users.noreply.github.com>


nemesifier (changes requested):
Thanks @phusi319, this is a real and well-scoped security fix and I appreciate the focused regression tests. The gap is genuine: `admin_view` only guarantees the user is active staff, it does not enforce model- or object-level permissions, so `json_view` / `floorplans_json_view` were returning location and floorplan data to any staff user regardless of their location permissions. Calling `has_view_or_change_permission(request, obj=instance)` and raising `PermissionDenied` is exactly the right pattern, and it correctly delegates to the multitenant admin override so organization scoping is respected too.

A couple of things to round it out before merge:

1. **Add a positive test so we do not over-restrict.** The two new tests confirm the deny path (403 without permission). Please also add the allow path: a user who does have location view permission still gets 200 and the expected JSON. That locks in that we did not accidentally break the normal flow, which matters because of point 2.
2. **Confirm the geo widget flow still works for operators.** These endpoints are consumed by `loci.js` when a user selects an existing location in the location/floorplan inline (for example when editing a device in openwisp-controller). After this change, that user must hold location view-or-change permission, otherwise the floorplan dropdown will start returning 403. I am fairly sure the standard openwisp operator groups already grant location permissions, so this should be fine, but please verify the controller's device-edit flow for a plain operator and call it out in the PR description. If any default group lacks location perms, that needs to be addressed alongside this.

The core change is correct and I want it merged. Just add the positive test and confirm the operator flow.

PR #158, opened 2025-03-20
[fix] Fixed ObjectLocation admin read-only mode #95

DragnEmperor:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [x] I have manually tested the changes proposed in this pull request.
- [x] I have written new test cases for new code and/or updated existing tests for changes to existing code.

## Reference to Existing Issue

Closes #95.

## Description of Changes

For view only permission, django admin uses 'AdminReadonlyField' for rendering data which looks for form fields in the model instance. And for widget to render 'read_only' must be set True. Setting the attributes to instance and 'read_only' geometry widget allowed the data to be viewed and map to render.

PR #230, opened 2026-06-06
[fix] Fixed ObjectLocation admin read-only mode on Django >= 6

### LINKED ISSUE #95: [bug] ObjectLocation admin doesn't work in readonly mode
When a user has only view permission, the map doesn't work because the JS fails (it was not tested in this scenario).

Dhanus3133:
## Checklist

- [ ] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [ ] I have updated the documentation.

Fixes #95

## Description of Changes

Django 6.0 removed the AdminReadonlyField.contents() path the previous #158 fix relied on, so the map/image widgets stopped rendering for view-only users. The fix re-adds it in AppConfig.ready(), only for widgets that set read_only=True.

Before:
image

After:
image

### DIFF
--- a/django_loci/apps.py
+++ b/django_loci/apps.py
@@ -42,7 +42,10 @@ class LociConfig(AppConfig):
     def ready(self):
         import leaflet
 
+        from .base.admin import restore_readonly_widget_rendering
+
         leaflet.app_settings["NO_GLOBALS"] = False
+        restore_readonly_widget_rendering()
         self.__setmodels__()
         self._load_receivers()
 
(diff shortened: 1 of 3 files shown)


nemesifier (commented):
Hey @Dhanus3133 what do you think about the possibility of helping us to push a fix upstream? Please see https://code.djangoproject.com/ticket/30577, we haven't head from the maintainers on the topic but maybe a well thought patch there could wake them up?

Nonetheless, as soon as I have time I'll try this fix, it could be a quick workaround to help us overcome this issue, but it's not a good long term solution.

